PR #382, opened 2015-01-12
Authorize municipality codes for premium user

mhautala:


src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -261,12 +268,25 @@ class Digiroad2Api extends ScalatraServlet with JacksonJsonSupport with CorsSupp
     }
   }
 
+  private def getMunicipalityCodes(assetId: Long): Set[Int] = {
+    val roadLinkIds = Database.forDataSource(ds).withDynTransaction {
```
uhef: Having direct sql queries on this level violates system abstraction leveling. I would split this to two operations: `getAssetRoadLinkIds` that would recide in `AssetProvider` and `getMunicipalityCode` that would recide in `RoadLinkService`

PR #413, opened 2015-03-16
Allow manoeuvres only on car traffic roads

mhautala:


UI/src/view/ManoeuvreLayer.js
```diff
@@ -71,9 +71,13 @@
...
       onSelect: function(feature) {
-        roadLayer.setLayerSpecificStyleMap(layerName, selectionStyleMap);
-        roadLayer.redraw();
-        selectedManoeuvreSource.open(feature.attributes.roadLinkId);
+        if (selectedManoeuvreSource.isCarTrafficRoad(feature.attributes.roadLinkId)) {
+          roadLayer.setLayerSpecificStyleMap(layerName, selectionStyleMap);
+          roadLayer.redraw();
+          selectedManoeuvreSource.open(feature.attributes.roadLinkId);
+        } else {
+          selectControl.unhighlight(feature);
```
uhef: Perhaps we should behave like we do in else branch (when manoeuvre is unselected) and call `unselectManoeuvre`.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/RoadLinkService.scala
```diff
@@ -313,19 +315,20 @@ object RoadLinkService {
         val boundingBoxFilter2 = OracleDatabase.boundingBoxFilter(bounds2)
 
         sql"""
-        select dr1_id, mml_id, to_2d(shape)
+        select dr1_id, mml_id, to_2d(shape), linkkityyppi
```
uhef: We can remove `linkkityyppi` from `select` - clause and add `where` filter to sql where we return only road links that are of car traffic - type. With this we can also remove filter at line 331.

PR #417, opened 2015-04-21
Use vvh geometry with mass transit stops

uhef:


UI/src/utils/backend-utils.js
```diff
@@ -197,22 +203,18 @@
...
       });
     };
 
-    this.getAssetWithCallback = function(assetId, callback) {
-      $.get('api/assets/' + assetId, callback);
-    };
-
-    this.getAssetByExternalId = function (externalId, callback) {
-      $.get('api/assets/' + externalId + '?externalId=true', callback);
+    this.getAssetByExternalId = function(nationalId, callback) {
```
mhautala: Maybe it would be better to rename 'getAssetByExternalId' as 'getAssetByNationalId'  
(or 'getMassTransitStopByNationalId')

PR #432, opened 2015-06-17
Adjust speed limit segments to road link lengths

uhef:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/oracle/OracleLinearAssetProvider.scala
```diff
@@ -109,17 +108,37 @@ class OracleLinearAssetProvider(eventbus: DigiroadEventBus, roadLinkServiceImple
...
     Database.forDataSource(ds).withDynTransaction {
-      logger.info("Filling partially filled road links, road link count in bounding box: " + linkGeometries.size)
-      OracleLinearAssetDao.fillPartiallyFilledRoadLinks(linkGeometries)
-      logger.info("...done with filling.")
+      val (speedLimitLinks, roadLinksByMmlId) = dao.getByMunicipality(municipality)
+
+      val (filledTopology, speedLimitChangeSet) = SpeedLimitFiller.fillTopology(roadLinksByMmlId, speedLimitLinks.groupBy(_.assetId))
+      eventbus.publish("speedLimits:update", speedLimitChangeSet)
+
+      filledTopology.map { link =>
```
uhef: Perhaps this transformation could be moved to integration api or whoever calls this function. Then getSpeedLimits by municipality and getSpeedLimits by bounding box would return similar data.

PR #456, opened 2015-08-19
Speed limit work list

uhef:


UI/src/application.js
```diff
@@ -35,6 +37,20 @@ var URLRouter = function(map, backend, models) {
         });
         map.setCenter(new OpenLayers.LonLat(response.middlePoint.x, response.middlePoint.y), 12);
       });
+    },
+
+    speedLimit: function(mmlId) {
+      applicationModel.selectLayer('speedLimit');
+      backend.getRoadLinkByMMLId(mmlId, function(response) {
+        eventbus.once('speedLimits:available', function() {
```
uhef: `speedLimits:available` is triggered from `applicationModel.selectLayer` once speed limits are fetched. If speed limit fetch completes before `getRoadLinkByMMLId` this event handler will never be triggered and the flow fails.

UI/src/less/site/global.less
```diff
@@ -134,3 +138,7 @@ a {
   color: @input-color-dark;
   text-transform: uppercase;
 }
+
+#work-list {
```
uhef: We now have both id `work-list` and class `work-list`. Should other be renamed?

UI/src/view/WorkListView.js
```diff
@@ -0,0 +1,69 @@
...
+        $('body').removeClass('work-list').scrollTop(0);
+      });
+    });
+  };
+
+  var bindEvents = function(backend) {
+    eventbus.on('workList:select', function(layerName) {
+      $('.container').hide();
+      $('#work-list').show();
+      $('body').addClass('work-list');
```
uhef: Do we need to add `work-list` class to body? Could it be added to one of the elements below `#work-list`?

UI/unknown_speed_limits.html
```diff
@@ -0,0 +1,70 @@
+<!DOCTYPE html>
```
uhef: Can this file be removed since we now have the embedded work list?

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/oracle/OracleLinearAssetDao.scala
```diff
@@ -20,6 +21,63 @@ import scala.util.Try
 
 
 trait OracleLinearAssetDao {
+  def getUnknownSpeedLimits(municipalities: Option[Set[Int]]): Map[String, Map[String, Seq[Long]]] = {
+    case class UnknownLimit(mmlId: Long, municipality: String, administrativeClass: String)
+    def toUnknownLimit(x: (Long, String, Int)) = UnknownLimit(x._1, x._2, AdministrativeClass(x._3).toString)
+    val optionalMunicipalities = municipalities.map(_.mkString(","))
+    val incompleteLinksQuery = """
```
uhef: Rename variable

PR #457, opened 2015-08-20
Unify work lists

mhautala:


UI/src/view/AssetForm.js
```diff
@@ -466,12 +466,27 @@
...
+            '</div>');
+        }
+      };
+
       eventbus.on('asset:modified', function(){
         renderAssetForm();
       });
 
-      eventbus.on('layer:selected', function(layer) {
-        activeLayer = layer;
```
uhef: `activeLayer` assignment is removed. Does this brake something? If not then we can remove `activeLayer` variable completely since it can never be anything else but `massTransitStop`

UI/src/view/AssetForm.js
```diff
@@ -466,12 +466,27 @@
...
+      eventbus.on('layer:selected application:initialized', function(layer) {
+        if(applicationModel.getSelectedLayer() === 'massTransitStop') {
```
uhef: `layer` should be same than what `applicationModel.getSelectedLayer()` returns. No need to requery the layer name from model.

PR #473, opened 2015-10-13
Cancel previous requests on concurrent requests

uhef:


wolverian:
Your code is good and you should feel good

mhautala:
Good shit

PR #481, opened 2015-12-03
Use vvh geometry for manoeuvres

simoraman:


UI/src/utils/backend-utils.js
```diff
@@ -10,15 +10,9 @@
         });
     };
 
-    this.getRoadLinks = createCallbackRequestor(function(boundingBox) {
-      return {
-        url: 'api/roadlinks?bbox=' + boundingBox
-      };
-    });
-
     this.getRoadLinksFromVVH = createCallbackRequestor(function(boundingBox) {
```
mhautala: Rename to simply getRoadlinks and get rid of the 'VVH' term in the front-end (which kind of a false information as the front-end doesn't communicate with the VVH api directly)

PR #494, opened 2016-02-05
Vvh here flip

mhautala:


tvenhola:
Point assets get end mValues (unused elsewhere), otherwise no issues. If later required, can be updated manually back to zero for flipped assets.

PR #503, opened 2016-04-05
Speed limit change history

tvenhola:
Code review, cleanup and refactoring time


tvenhola:
I'll generify some of the speed limit related stuff that could go to linear asset operations.

PR #504, opened 2016-04-18
Return also removed speed limits from change api

mhautala:
Speed limits are now removed when the road link they are on is no more valid in VVH. Return these removed speed limits from change api.


tvenhola:
Looks good to me.

leirasf:
Looks good

PR #505, opened 2016-04-18
Speed limit filler improvement

tvenhola:
Improve filler so that it automagically fixes the errors that may appear or have been entered in the database and that it is stable (not flip-flopping)


tvenhola:
We should use the Epsilon in comparison where applicable: on floating point arithmetic the rounding errors may affect the equality comparison when they are coming from different sources.

PR #511, opened 2016-05-04
Filter geometry adjustements from change api

mhautala:
The TN-ITS API is the only client for the change api and it does not care about the changes propagating from the changes to the VVH roadlinks. Including changes to geometry and link id's.

Hence simply filter out all these adjustements.


tvenhola:
Looks good to me.

PR #512, opened 2016-05-10
Fix multiple linear assets appearing on same road link on top of each…

tvenhola:
… other and expiring not propagating from filler to service output. Includes a test with data imported from staging.


digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/NumericalLimitFiller.scala
```diff
@@ -171,10 +174,75 @@ object NumericalLimitFiller {
...
+    val keeperOpt = sorted.headOption
+    if (keeperOpt.nonEmpty) {
+      val keeper = keeperOpt.get
+      val (overlapping) = sorted.tail.flatMap(asset => GeometryUtils.overlap(toSegment(keeper), toSegment(asset)) match {
+        case Some(overlap) =>
+          if (keeper.sideCode == asset.sideCode || keeper.sideCode == SideCode.BothDirections.value) {
+            Seq(
+              asset.copy(startMeasure = asset.startMeasure, endMeasure = overlap._1),
+              asset.copy(id = 0L, startMeasure = overlap._2, endMeasure = asset.endMeasure)
+            ).filter(a => a.endMeasure - a.startMeasure >= AllowedTolerance)
```
leirasf: After spliting the overlaped asset, if both are greater than AllowedTolerance they will be both filtered, even if it's not the ovelaped part. This is the expected result?

leirasf: Maybe we should change the name of the method "sortAndDrop" because he doesn't performe the sort. He just keeps the order of the sequence "sorted"

tvenhola: Too short pieces are dropped intentionally: the tolerance is 50 cm.

halied:
Ok

JaniKoivisto:
ok

PR #513, opened 2016-05-12
Linear asset change history

tvenhola:


halied:
Are those prohibitions TODO comments there on purpose?

PR #514, opened 2016-05-24
Linear asset change history

tvenhola:
Improvements: 
- road link extensions should also extend the assets
- thruput
- add vvh timestamp (create a pseudo one) for newly created assets and speed limits as they are up-to-date with the geometry already

### DIFF
--- a/digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/GeometryUtils.scala
+++ b/digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/GeometryUtils.scala
@@ -210,5 +210,20 @@ object GeometryUtils {
     ((projection.oldEnd - projection.oldStart)*(projection.newEnd - projection.newStart)) < 0
   }
 
+  /**
+    * Creates a pseudo VVH Timestamp for new assets and speed limits. Turns clock back to 0:00 on the same day
+    * if less than offsetHours have passed since or 0:00 on previous day if not.
+    * @param offsetHours Number of hours since midnight to return current day as a VVH timestamp (UNIX time in ms)
+    */
+  def createVVHTimeStamp(offsetHours: Int): Long = {
+    val oneHourInMs = 60 * 60 * 1000L
+    val curr = System.currentTimeMillis
+    val offset = curr % 86400000L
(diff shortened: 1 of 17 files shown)


digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/GeometryUtils.scala
```diff
@@ -194,5 +215,18 @@ object GeometryUtils {
...
+    *
+    * @param offsetHours Number of hours since midnight to return current day as a VVH timestamp (UNIX time in ms)
+    */
+  def createVVHTimeStamp(offsetHours: Int): Long = {
+    val oneHourInMs = 60 * 60 * 1000L
+    val utcTime = DateTime.now().minusHours(offsetHours).getMillis
+    val curr = utcTime + DateTimeZone.getDefault.getOffset(utcTime)
+    curr - (curr % (24L*oneHourInMs))
+  }
+
```
leirasf: Maybe it's a good idea change the method createVVHTimeStamp to the VVClient instead of be in GeometryUtils

halied:
Might be off topic (not sure if these codes were in the pull request) but is it ok that in some places in the code vvhTimeStamp possibly gets  value 0?
For example in LinearAssetService row 244:
          case true => Some(Projection(oldStart, oldEnd, newStart, newEnd, vvhTimeStamp.getOrElse(0L)))
And OracleLinearAssetDao.scala row838:
vvhTimeStamp: Long = 0L 

And should vvhTimeStamp be Long instead of Option[Long] in VvhClient.scala ChangeInfo case class?

Is there still a need for default value 0 for adjusted_timestamp in db?

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/GeometryUtils.scala
tvenhola: Moved.

tvenhola:
@halied We should have a default value for db column at least as long as we don't have change info applied to all assets.
VVHTimeStamp should be on all change info rows, so removing the Option out of it and giving a fallback value of 0L just in case.

PR #522, opened 2016-07-28
Droth 239 linear asset projection bug

tvenhola:


tvenhola:
Fixes a linear asset projection bug: A projected asset that does not fill the whole segment is filled by end user; it was then re-projected.

PR #523, opened 2016-07-29
Droth 240 address translation

tvenhola:
Address translation functions from coordinates to road network address.


leirasf:
It looks very good to me

marquesfm:
Good to me, can proceed

PR #524, opened 2016-08-03
144

oappicgi:
DROTH-144 Subtask 37

User should not be able to choose virtualstop type with any other stop type.
Includes both UI and backend validations.


tvenhola:
Looks ok to me.

leirasf:
It looks good to me too

marquesfm:
Looks ok to me too.

PR #526, opened 2016-08-04
Droth 220 adding address field stop form

marquesfm:
Creation of the fields in the Data Base to show Finnish and Swedish address;
Write the correct order to show this two Fields.


tvenhola:
It might be that we need to include this to integration API - but let's not go there just yet.

tvenhola:
Otherwise, looks ok to me.

leirasf:
Looks good to me too

PR #527, opened 2016-08-16
DROTH-247 New manoeuvres bug about moving the map during manoeuvre creation

marquesfm:
- Corrections at reselectManoeuvre and in getAdjacents;


halied:
Looks ok

leirasf:
Looks good to me too

PR #542, opened 2016-09-27
Drvvh 209

tvenhola:


tvenhola:
Track code 0 does not show on form

tvenhola:
Fixed showing of missing values, ok for me now.

JaniKoivisto:
According to userstory do we need "staticField('SEGMENTIN ID', 'segmentId')" in form? Otherwise looks good!

tvenhola:
Segment id is a good addition, remove it later if necessary. Now it clears up some confusion there might be.

JaniKoivisto:
Ok

PR #545, opened 2016-10-03
Drvvh 288 initial load m value adjustments

tvenhola:
Import batch run improvements, a filler starting point and address (re-)calculator for road parts.


pedrosag:
Seems all right but the TODO at line 35 of LinkRoadAddressCalculator.scala is it done or is it a task for another time, also the comment at line 43 in the same file is it already done, is it ok to remove?

JaniKoivisto:
Ok

tvenhola:
TODO is when the link geometry (VVH ChangeInfo) updates are affecting (joining, separating) links. It is not applicable to DRVVH-294 according to PO.

PR #547, opened 2016-10-04
DRVVH-300 Text "Yleinen tie" on the form will be changed to word "Maantie" and "Omistaja selvittämättä" to "Ei tiedossa"

pedrosag:
Changed the texts.


tvenhola:
Trivial; will merge.

PR #548, opened 2016-10-04
Drvvh 294 recalculate after initial mValue import

tvenhola:
A batch job to recalculate mValues after import task.


pedrosag:
Seems ok

JaniKoivisto:
Ok

PR #549, opened 2016-10-04
Droth294

oappicgi:
Frontend asks google street views image for mass transit stops from backend with coordinates and heading parameters.  Backend returns image address with signed parameter. 

Backend signs requests with crypto-key given by google. If key with client-id has not been added to /digiroad2/conf/*folder*/key.properties file, backend will fallback to old api-key.

Required attributes to be added into key.properties file:
googlemapapi.client_id=
googlemapapi.crypto_key=


leirasf:
It looks good to me

marquesfm:
Seems ok to me

tvenhola:
Will merge.

PR #550, opened 2016-10-04
DRVVH-299 Closure date (LAKKAUTUS) on the form has to be <null>, nothing should be shown on the form there

pedrosag:
Altered the from creation to mirror the following behaviour:
If the field LAKKAUTUS value would have no value then it's field will not be included in the form.


JaniKoivisto:
Should the text 'LAKKAUTUS' still be visible even if we dont have value?

Campanha:
seems fine

pedrosag:
@JaniKoivisto 
Regarding your question about the visibility of the LAKKAUTUS field, from the user story I gathered that it was a field to be removed/hidden from the Form.

I asked on skype if the field would be removed would it be Ok in the future. I was told currently the Viite database does not have that data but the PO hinted that it might have in the future, therefore I made it so that it hides the field if there is no data.

leirasf:
Looks good to me

PR #551, opened 2016-10-06
Drvvh 258 identifying gap locations

Campanha:


viite-UI/src/view/LinkPropertyLayerStyles.js
```diff
@@ -1,23 +1,23 @@
...
   root.LinkPropertyLayerStyles = function(roadLayer) {
-    /*
-    / Todo: Handle unknown, stroke black with question mark
-    /
+
+    // Todo: Handle unknown, stroke black with question mark
+
     var unknownRoadClassDefaultRules = [
-      new OpenLayersRule().where('roadClass').is('99').use({ strokeColor: '#000000', strokeOpacity: 1.0, externalGraphic: 'images/speed-limits/unknown.svg', pointRadius: 14})
+      new OpenLayersRule().where('type').is('roadAddressAnomaly').use({ strokeColor: '#000000', strokeOpacity: 0.8, externalGraphic: 'viite-UI/images/speed-limits/unknown.svg', pointRadius: 14})
```
leirasf: I belive we could have the same image on a different folder instead of using speedlimit.

tvenhola:
Does not compile,
fixture reset does not run.

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/RoadAddressServiceSpec.scala
```diff
@@ -89,27 +89,28 @@ class RoadAddressServiceSpec extends FunSuite with Matchers{
...
         RoadAddressLink(0, 1611617, Seq(Point(374668.195, 6676884.282, 24.48399999999674), Point(374690.403, 6676887.31, 24.645000000004075)), 139.0852126853, State, 3, TrafficDirection.BothDirections, SingleCarriageway, Some("22.09.2016 14:52:19"), Some("dr1_conversion"), Map("linkId" -> 1611601, "segmentId" -> 63298), 0, 0, 0, 0, 0, 0, 0, "", 0.0, 0.0, SideCode.Unknown, None, None),
         RoadAddressLink(0, 1611618, Seq(Point(374668.195, 6676884.282, 24.48399999999674), Point(374643.384, 6676882.176, 24.42399999999907)), 297.7533188814259, State, 3, TrafficDirection.BothDirections, SingleCarriageway, Some("22.09.2016 14:51:28"), Some("dr1_conversion"), Map("linkId" -> 1611605, "segmentId" -> 63298), 0, 0, 0, 0, 0, 0, 0, "", 0.0, 0.0, SideCode.Unknown, None, None)
       )
-      roadAddressLinks.map { links =>
-          RoadAddressDAO.createMissingRoadAddress(links.linkId, links.startAddressM, links.endAddressM, links.roadNumber, links.roadPartNumber, 1)
+      roadAddressLinks.foreach { links =>
+        RoadAddressDAO.createMissingRoadAddress(links.linkId, links.startAddressM, links.endAddressM, links.roadNumber, links.roadPartNumber, 1)
       }
+      //TODO: Validation
```
tvenhola: Needs validation for results here - just not crashing when writing is not enough.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -79,7 +80,15 @@ class RoadAddressService(roadLinkService: RoadLinkService) {
...
       val ra = addresses.getOrElse(rl.linkId, Seq())
       buildRoadAddressLink(rl, ra)
+  }
+    val newAnomalousRL = viiteRoadLinks.filter(rl => (rl.administrativeClass == State || rl.roadNumber > 0) && rl.id == 0)
+
+    val missedRL  = withDynTransaction {
+    RoadAddressDAO.getMissingRoadAddresses()
     }
+
+    val missingRL = newAnomalousRL.filterNot(rl => missedRL.exists(mrl => mrl._1 == rl.linkId && mrl._2 == rl.startAddressM && mrl._3 == rl.endAddressM))
```
leirasf: On this case we are just looking at the road address that doesn't exist on the database, so the anomaly will be always code = 1. Shouldn't we just set the code to one on this case?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -79,7 +80,15 @@ class RoadAddressService(roadLinkService: RoadLinkService) {
     val viiteRoadLinks = roadLinks.flatMap { rl =>
       val ra = addresses.getOrElse(rl.linkId, Seq())
       buildRoadAddressLink(rl, ra)
+  }
+    val newAnomalousRL = viiteRoadLinks.filter(rl => (rl.administrativeClass == State || rl.roadNumber > 0) && rl.id == 0)
+
+    val missedRL  = withDynTransaction {
+    RoadAddressDAO.getMissingRoadAddresses()
```
leirasf: 1. I think is better to get only the roadAddresses with link id inside the viiteRoadLinks. Instead of getting the all table.
2. I belive this is not releated with this task but the RoadAddressDAO shouldn't be a class so we can mock it on the unit tests?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -203,4 +212,33 @@ class RoadAddressService(roadLinkService: RoadLinkService) {
...
+  }
+
+  def getAnomalyCodeByLinkId(linkId: Long, roadPart: Long) : Int = {
+    var code = 0
+      //roadlink dont have road address
+      if (RoadAddressDAO.getLrmPositionByLinkId(linkId).length < 1) {
+        code = 1
+      }
+      //road address do not cover the whole length of the link
+       else if (RoadAddressDAO.getLrmPositionMeasures(linkId).length > 0)
```
leirasf: This verification shouldn't be done with the lenght of the road geometry and the length of the road address? Instead of comparing the start and end measure of the LrmPosition?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
tvenhola: 1. We should rework this operation, maybe introduce a filler here (because later we will be using that way of working). A filler stub is already in the master branch.

2. I don't think it's beneficial - we anyway do the service testing on integration level. So let's just use a rollback operations around them.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
tvenhola: We should, as the future is not clear in this operation.

tvenhola:
I'll start merging, let's fix the issues in service after changes from master have been merged in.

tvenhola:
Needs still work to do: let's plug in the filler and use that for generating unknown road addresses list.
RoadAddressFiller.scala:48
RoadAddressService.scala:96

Let's use Filler to return us the new missing addresses in address change set. Already known ones can be returned as ordinary roadlinkaddresses with zeros in them.

tvenhola:
When anomalous road links have proper signs this is ready.

PR #552, opened 2016-10-06
DRVVH-301  Selecting "Muu tieverkko": there is no difference if clicking once or using double-click

pedrosag:
Changed the Layer in order to highlight all the features if single clicked, and if double clicked to highlight the single one.


JaniKoivisto:
Looks good

leirasf:
Looks good to me

PR #555, opened 2016-10-10
DROTH-312 (144) Bug: Bus stop removal is not implemented according to specification

marquesfm:
These features are missing from bus stop removal implementation (DROTH-273):
- There should be a confirmation popup for user when Poista checkbox is checked and user clicks Tallenna. See DROTH-273 for message text.
- The bus stop should be first removed from Tierekisteri and if that is successful, it can be removed from OTH
- If removal from Tierekisteri fails, user should be shown a popup message. See DROTH-273 for message text.


halied:
Unit tests missing, otherwise ok.

marquesfm:
Unit tests added.

PR #557, opened 2016-10-11
DRVVH-308 Form disappears unexpectedly

pedrosag:
Fixed a bug in the drawing calibration markers.

From the first screenshot it is shown that the sometimes the calibration markers are being made without coordinates.

I made a filter on Layer.js in order to only let the ones with coordinates pass to be drawn.

After making this fix the error i got disappeared and I am not able to reproduce the issue.

### DIFF
--- a/viite-UI/src/view/Layer.js
+++ b/viite-UI/src/view/Layer.js
@@ -90,7 +90,10 @@
       }).map(function(roadLink) {
         return roadLink.calibrationPoints;
       }));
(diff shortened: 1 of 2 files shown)


JaniKoivisto:
Ok. Should we prevent the user from clicking the point?

pedrosag:
Small typo on the 8edfe3b commit message, when reading "...preform an action.", should be read "...preform no action.".

PR #558, opened 2016-10-12
DRVVH-315 Drawing technique is broken. Selected road sections and segments are not drawn correctly

pedrosag:
Changed the LinkPropertyLayer in order to send it's style rulings to the RoadLayer, also made it to redraw on selection.

It now applies the correct layer rules to the highlighting,


tvenhola:
OK to me.

PR #560, opened 2016-10-12
DRVVH-302 - Filed Segmentin-ID needs to be removed from the lower section of the form

pedrosag:
Removed the segmentin ID field from the form, except in the header.


tvenhola:
Small, merged to master branch.

PR #561, opened 2016-10-12
DROTH-317 Check tierekisteri equipment if it's master

leirasf:
When Tierekisteri data fetch for bus stop form was implemented, it was not known which equipment fields are stored in Tierekisteri and which in OTH. 
These are the five equipment fields that have Tierekisteri as a master db: aikataulu, roskis, pyorateline, valaistus and penkki. 
It should be checked that these are the only fields that come to bus stop object from Tierekisteri and the others are fetched from OTH db.


marquesfm:
Ok for me.

halied:
I can't get unit tests passed on my local environment, otherwise ok by me.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/MassTransitStopService.scala
```diff
@@ -102,9 +102,11 @@ trait MassTransitStopService extends PointAssetOperations {
       return property
     val equipmentProperty = tierekisteriStop.equipments.filter(_._1.publicId == property.publicId)
     val mappedProperties = equipmentProperty.map {
-      case (equipment, existence) =>
-        PropertyValue(existence.propertyValue.toString, Some(massTransitStopDao.getPropertyDescription(property.publicId, existence.propertyValue.toString).head))
-    }
+      case (equipment, existence) if(equipment.isMaster) =>
+        Some(PropertyValue(existence.propertyValue.toString, Some(massTransitStopDao.getPropertyDescription(property.publicId, existence.propertyValue.toString).head)))
```
tvenhola: We should avoid calling DAO in a loop, too much connection overhead, especially in batch runs.

tvenhola:
Looks good to me. Any other takers?

leirasf:
It look's good to me

PR #562, opened 2016-10-13
DRVVH-319 If the road section selected has several road types, they will be listed sequentially in field "Tietyyppi"

pedrosag:
Changed the form in order to concatenate all the unique roadClasses separated with ','


tvenhola:
Small, merged right away.

PR #565, opened 2016-10-13
DROTH-316 (144) Bug: Form is not opened for existing bus stop if there is no bus stop with the same LiviId in Tierekisteri

marquesfm:
- Corrections at Fetch MassTransitSops to always return, and in case can't comunicate with TR send a pop-up with an alert;
- Create new Error to be send when MassTransitStop don't exist in TR;
- Corrections and creations of new unit tests.


halied:
Ok by me

tvenhola:
Will merge

PR #566, opened 2016-10-17
DRVVH-318 The form should not have OTH modifiedby information

Campanha:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/VVHClient.scala
```diff
@@ -514,9 +514,11 @@ class VVHClient(vvhRestApiEndPoint: String) {
       "MTKHEREFLIP",
       "VALIDFROM",
       "GEOMETRY_EDITED_DATE",
+      "CREATED_DATE",
+      "LAST_EDITED_DATE",
       "SURFACETYPE").contains(x)
-    }.filter { case (_, value) =>
-      value != null
+    }.filter { case (key, value) =>
+      value != null || key == "LAST_EDITED_DATE" || key == "GEOMETRY_EDITED_DATE"
```
Campanha: to let the properties "LAST_EDITED_DATE" and "GEOMETRY_EDITED_DATE" go to extractedProperties, to be treated after in RoadAddressService

tvenhola: But if their values are null, is there any difference in logic compared to the fact that if the key is missing the value must be null, too?

tvenhola:
This is ok

PR #569, opened 2016-10-18
DROTH-328 (144) Administrator change from ELY-keskus to Kunta not wor…

leirasf:
At the moment when bus stop administrator is changed from ELY-keskus to something else, OTH tries to call Tierekisteri PUT endpoint without Livi-Id.

### DIFF
--- a/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/MassTransitStopService.scala
+++ b/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/MassTransitStopService.scala
@@ -426,7 +426,13 @@ trait MassTransitStopService extends PointAssetOperations {
         case (false, true) => Operation.Create
         case (false, false) => Operation.Noop
       }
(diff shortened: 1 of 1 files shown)


marquesfm:
Ok for me.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/MassTransitStopService.scala
```diff
@@ -513,20 +519,31 @@ trait MassTransitStopService extends PointAssetOperations {
...
 
   private def getPersistedStopWithPropertiesAndPublishEvent(assetId: Long, roadLinkByLinkId: Long => Option[VVHRoadlink],
-                                                            operation: Operation,
-                                                            relevantToTR: Boolean = false) = {
+                                                            operation: Operation = Operation.Noop) = {
     val persistedStop = fetchPointAssets(withId(assetId)).headOption
 
-    if (relevantToTR) {
-      val roadLink = roadLinkByLinkId.apply(persistedStop.get.linkId)
+    persistedStop.foreach { stop =>
```
halied: Why does this have foreach? Isn't there only one asset returned with headOption?

halied:
Added some comments, otherwise ok

PR #570, opened 2016-10-18
DRVVH-277 Identifying floating road address segments and saving them to the DB

pedrosag:
Added the evocation of the batch on the Viite Data Fixture, to evoke it use "find_floating_road_addresses" the amount of fetched rows can be changed by altering the "batchSize" variable;

Added the sql queries in RoadAddressDAO in order to consult the amount of road addresses existent, to fetch a portion of said road addresses and to update the floating field of the road address.


tvenhola:
Okay to me.

JaniKoivisto:
Ok

PR #572, opened 2016-10-20
Drvvh 334 ramps should only be shown on closer zoom levels

Campanha:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/VVHClient.scala
```diff
@@ -385,7 +385,7 @@ class VVHClient(vvhRestApiEndPoint: String) {
     * Used by RoadAddressService.getMissingRoadAddresses
     */
   def fetchByMunicipality(municipality: Int, roadNumbers: Seq[(Int, Int)]): Seq[VVHRoadlink] = {
-    val roadNumberFilters = withRoadNumbersFilter(roadNumbers, true, "")
+    val roadNumberFilters = withRoadNumbersFilter(roadNumbers, false, "")
```
tvenhola: On municipality search we are searching for missing addresses -> changing this to false breaks missing address search.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/RoadLinkService.scala
```diff
@@ -298,7 +298,7 @@ class RoadLinkService(val vvhClient: VVHClient, val eventbus: DigiroadEventBus,
     */
   def getViiteRoadLinksAndChangesFromVVH(bounds: BoundingRectangle, roadNumbers: Seq[(Int, Int)],
                                          municipalities: Set[Int] = Set(), everything: Boolean,
-                                         publicRoads: Boolean): (Seq[RoadLink], Seq[ChangeInfo])= {
+                                         publicRoads: Boolean = false): (Seq[RoadLink], Seq[ChangeInfo])= {
```
tvenhola: We're probably better of with code readability if these service level options don't have default values or at least these two either both have a default value or neither does.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/RoadLinkService.scala
```diff
@@ -349,7 +349,7 @@ class RoadLinkService(val vvhClient: VVHClient, val eventbus: DigiroadEventBus,
     getRoadLinksAndChangesFromVVH(bounds, municipalities)._1
 
   def getViiteRoadLinksFromVVH(bounds: BoundingRectangle, roadNumbers: Seq[(Int, Int)], municipalities: Set[Int] = Set(),
-                               everything: Boolean, publicRoads: Boolean) : Seq[RoadLink] =
+                               everything: Boolean, publicRoads: Boolean = false) : Seq[RoadLink] =
```
tvenhola: We're probably better of with code readability if these service level options don't have default values or at least these two either both have a default value or neither does.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/VVHClient.scala
```diff
@@ -169,7 +169,7 @@ class VVHClient(vvhRestApiEndPoint: String) {
     withLimitFilter("ROADNUMBER", roadNumbers._1, roadNumbers._2, includeAllPublicRoads)
   }
 
-  private def withRoadNumbersFilter(roadNumbers: Seq[(Int, Int)], includeAllPublicRoads: Boolean, filter: String = ""): String = {
+  private def withRoadNumbersFilter(roadNumbers: Seq[(Int, Int)], includeAllPublicRoads: Boolean = false, filter: String = ""): String = {
```
tvenhola: I think it's better to not have a default value here: it's internal and whether the boolean is set to true or false has a severe impact on the functionality as a whole. If it was for compatibility (and it wasn't private def) then it'd be good to have it here.

PR #577, opened 2016-10-25
Drvvh 312 creating log of the initial data load

leirasf:
A log needs to be created of all the rows that cannot be saved to Viite during the initial data load. All attribute data needs to be picked up. Reason codes used:
1 = Link-ID is not found in the VVH interface
2 = Values of the start and end fields are totally outside of the link geometry 
3 = Another reason
These are for example cases where the link-ID is not found in the VVH interface and values of the start and end fields of a road address segment are totally outside of the link geometry (for example the length of the link is 10, road address start is 12 and end 18).


tvenhola:
This will wait until OTH deployment is done

tvenhola:
Looks good.

PR #580, opened 2016-10-28
Droth 331 Authentication

oappicgi:
*Changed authentication to different header attribute. 
*Adds keys.properties file where  TR username and password should be added
( googlemapapi fallsback to old key  if it cannot find crypto-key/client-id combo, so it can be left commented out).


leirasf:
It's ok to me

PR #581, opened 2016-10-28
DROTH-291 Batch run to update stops that are missing in the road register

marquesfm:
- Create Batch for verify if Bus Stop existed at OTH exist too at Tierekisteri, if doesn't exist send a request to Tierekisteri for create the Bus Stop.


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -479,6 +500,61 @@ object DataFixture {
...
+
+    municipalities.foreach { municipality =>
+      println("Start processing municipality %d".format(municipality))
+
+      //Get all OTH Bus Stops By Municipality
+      persistedStop = massTransitStopService.getByMunicipality(municipality, true)
+
+      persistedStop.foreach { stop =>
+        // Validate if OTH stop are known in Tierekisteri and if is maintained by ELY
+        if (!liviIdsListTR.exists(_ == "OTHJ%d".format(stop.nationalId)) && massTransitStopService.isStoredInTierekisteri(Some(stop))) {
```
tvenhola: Not all stops are of that form - some stops have older LiviIDs which are saved in stops' properties.

halied:
Should we have separate batch processes for the listing and the update?

tvenhola:
Maybe some kind of dry-run option would be good for trying it out against the server data.

tvenhola:
Looks good to me

PR #582, opened 2016-10-28
Drvvh 114 rebased viite integration api

leirasf:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -235,10 +235,46 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
     }
   }
 
+  /*
+    Kalpa-API methods
+  */
+
+  def getRoadAddressesLinkByMunicipality(municipality: Int): Seq[RoadAddressLink] = {
+    val linkIdsAmmounts = 100
+    var roadLinks = roadLinkService.getViiteRoadLinksFromVVHByMunicipality(municipality)
```
leirasf: Avoid the use of the "var" we can have one val for roadlinks and another for filtered roadlinks

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -285,10 +321,10 @@ object RoadAddressLinkBuilder {
...
     val length = GeometryUtils.geometryLength(geom)
     new RoadAddressLink(roadAddress.id, roadLink.linkId, geom,
-      length, roadLink.administrativeClass, roadLink.linkType, getRoadType(roadLink.administrativeClass, roadLink.linkType),
-      extractModifiedAtVVH(roadLink.attributes), Some("vvh_modified"),
+      length, roadLink.administrativeClass,
+      roadLink.linkType, getRoadType(roadLink.administrativeClass, roadLink.linkType), extractModifiedAtVVH(roadLink.attributes), Some("vvh_modified"),
       roadLink.attributes, roadAddress.roadNumber, roadAddress.roadPartNumber, roadAddress.track.value, municipalityMapping.getOrElse(roadLink.municipalityCode, -1), roadAddress.discontinuity.value,
-      roadAddress.startAddrMValue, roadAddress.endAddrMValue, roadAddress.endDate.map(formatter.print).getOrElse(""), roadAddress.startMValue, roadAddress.endMValue,
+      roadAddress.startAddrMValue, roadAddress.endAddrMValue, roadAddress.startDate.map(formatter.print).getOrElse(null), roadAddress.endDate.map(formatter.print).getOrElse(""), roadAddress.startMValue, roadAddress.endMValue,
```
leirasf: Should getOrElse a empty string instead of a null

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -235,10 +235,46 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
+  def getRoadAddressesLinkByMunicipality(municipality: Int): Seq[RoadAddressLink] = {
+    val linkIdsAmmounts = 100
+    var roadLinks = roadLinkService.getViiteRoadLinksFromVVHByMunicipality(municipality)
+    val linkIds = roadLinks.map(_.linkId).toSet
+    val linkIdGroups = linkIds.grouped(linkIdsAmmounts)
+
+    val addresses = linkIdGroups.map(links =>
+      withDynTransaction {
+      RoadAddressDAO.fetchByLinkId(links)
+    }).flatten.toList.groupBy(_.linkId)
```
tvenhola: We should get those links in one transaction if possible and with one query if possible. So not for each group but get them all from the database at once and then use map to combine the database loaded items with the link id groups.

src/main/scala/ScalatraBootstrap.scala
```diff
@@ -22,6 +23,7 @@ class ScalatraBootstrap extends LifeCycle {
     context.mount(new PingApi, "/api/ping/*")
     context.mount(new MassTransitStopImportApi, "/api/import/*")
     context.mount(new IntegrationApi(Digiroad2Context.massTransitStopService), "/api/integration/*")
+    context.mount(new ViiteIntegrationApi(Digiroad2Context.roadAddressService), "/viite/api/integration/*")
```
tvenhola: Is this the correct path?

src/main/scala/fi/liikennevirasto/digiroad2/ViiteIntegrationApi.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  before() {
+    basicAuth
+  }
+
+  def geometryWKT(geometry: Seq[Point]): (String, String) = {
+    if (geometry.nonEmpty) {
+      val segments = geometry.zip(geometry.tail)
+      val runningSum = segments.scanLeft(0.0)((current, points) => current + points._1.distance2DTo(points._2))
```
tvenhola: M value not according to definition (start = start address M value)

src/main/scala/ScalatraBootstrap.scala
leirasf: The local url to test is "http://localhost:8080/viite/api/integration/road_address?municipality=235" so it seems to be correct.

tvenhola: Viite APIs go under /api/viite/

tvenhola:
Okay for me

PR #583, opened 2016-10-31
Drvvh 357 remove ely table and to the municipality table

pedrosag:
Created a script in order to remove ELY table out of the database.
Before elimination all data is mapped onto 1 of 2 new columns of the municipality table.
Purged all reference of ELY table and replaced it with the correct calling on the municipality table.


pedrosag:
DRVVH-362 ELY-code to the form from the ROAD_MAINTAINER_ID column of the MUNICIPALITY-table is also included in this pull request.

pedrosag:
Making some test cases for the MunicipalityDAO.

tvenhola:
OK for me

PR #584, opened 2016-11-01
Drvvh 265 create modules for deployment

tvenhola:
Split modules according to new design to enable independent building of OTH and Viite modules.


pedrosag:
Seems fine except that the following test is breaking:

[info] OracleUserProviderSpec:
...
[info] - Delete all MassTransitStop data by Id  FAILED 
[info]   java.sql.SQLException: ORA-00060: deadlock detected while waiting for resource

PR #588, opened 2016-11-11
Droth 351 bus stop maintainer role not working without operator role

marquesfm:
Give privileges to busStopMaintainer role like operator role.


halied:
Ok

leirasf:
At least for now it's ok.

PR #589, opened 2016-11-17
Droth 337 use cached geometry

tvenhola:
Use RoadLinkService instead of raw VVHClient in point asset operations and all point assets. Use cached geometry if requesting full municipality data.


leirasf:
It ok to me.

marquesfm:
Ok for me.

halied:
Ok

oappicgi:
Ok

PR #592, opened 2016-11-23
Drvvh 278 showing floating road address links

tvenhola:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/RoadLinkService.scala
```diff
@@ -428,6 +429,24 @@ class RoadLinkService(val vvhClient: VVHClient, val eventbus: DigiroadEventBus,
...
   }
 
+  def getViiteRoadLinksHistoryFromVVH(roadAddressesLinkIds: Set[Long]): Seq[VVHHistoryRoadLink] = {
+    if (roadAddressesLinkIds.nonEmpty) {
+      val historyData = Await.result(vvhClient.historyData.fetchVVHRoadlinkHistoryF(roadAddressesLinkIds), atMost = Duration.Inf)
+      val groupedData = historyData.groupBy(_.linkId)
+      groupedData.mapValues(_.maxBy(_.endDate)).values.toSeq
+    } else
+      Nil
+  }
```
leirasf: It shouldn't return a None and then use a Option[Seq[VVHHistoryRoadLink]]?

PR #604, opened 2017-01-02
Drvvh 467 style change for muu tieverkko rakenteilla

JaniKoivisto:


viite-UI/src/view/LinkPropertyLayer.js
```diff
@@ -275,7 +280,7 @@
       if (linkPropertiesModel.getDataset() === 'functional-class') {
```
tvenhola: We could do some cleanup here. No data sets in use.

tvenhola:
Some cleanup would be nice, otherwise ok.

pedrosag:
@tvenhola I made the cleanup, the rest look a-ok.

PR #605, opened 2017-01-02
Droth 40 search by

oappicgi:
Feature to search with link-id, masstransitstop id (national & livi-id) and speedlimit id.


halied:
OK

UI/src/router.js
```diff
@@ -50,20 +50,30 @@
...
+                eventbus.once('linkProperties:available', function () {
+                  models.selectedLinkProperty.open(response.id);
+                });
+              });
+            }
+            map.setCenter(new OpenLayers.LonLat(response.middlePoint.x, response.middlePoint.y), 12);
+          }
+          else
+          {
+            //might be nice to show error message for user if roadlink  applied to #linkProperty/ url does not exist
```
marquesfm: I agree with the idea of the show error message for user if roadlink doesn't exist.
If we don't have the message yet, maybe it's better make this comment like a TODO.

joanacoelho:
I agree with the previous comment, but everything else seems ok.

marquesfm:
With the exception of the comments is all OK for me.

UI/src/model/LocationSearch.js
```diff
@@ -14,30 +22,176 @@
...
+     * @returns {*}
+     */
+    var roadNumberAndSpeedLimitSearch = function(input){
+      var roadNumberSearch = backend.getCoordinatesFromRoadAddress(input);
+      var speedlimitSearch= backend.getSpeedLimitsLinkIDFromSegmentID(input);
+      return $.when(
+        speedlimitSearch, roadNumberSearch).then(function(speedlimitdata,roadData) {
+        var returnObject = roadLocationAPIResultParser(roadData);
+        var linkFound =_.get(speedlimitdata[0], 'success');
+        if (linkFound === true) {
```
MAC-CGI: In this case I thing that you can remove the variable 'link Found' and put directly the content in the if clause, like you have done in line 81 of this file. Do you agree?

MAC-CGI:
In addition of my suggestion, everything seems ok.

UI/src/model/LocationSearch.js
oappicgi: Yea, it is probably clear enough that way. Thank you

PR #606, opened 2017-01-05
Droth 355 kalpa fix

oappicgi:
Kalpa-API fix for 416 that fixes:
Unknown link-sources for all links,
Missing complementary links


tvenhola:
Railroading this to the master.

PR #607, opened 2017-01-06
Droth 43 prevent adding bus stops agains traffic direction

marquesfm:


UI/src/view/MassTransitStopLayer.js
```diff
@@ -354,14 +354,15 @@ window.MassTransitStopLayer = function(map, roadCollection, mapOverlay, assetGro
   };
 
   var createNewAsset = function(lonlat, placement) {
+    var directions_decode = {BothDirections: 2, TowardsDigitizing: 2, AgainstDigitizing: 3};
```
tvenhola: Should be probably
`BothDirections: 1`

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -647,6 +647,62 @@ object DataFixture {
...
+    municipalities.foreach { municipality =>
+      println("Start processing municipality %d".format(municipality))
+
+      //Get all OTH Bus Stops By Municipality
+      persistedStop = massTransitStopService.getByMunicipality(municipality, false)
+
+      persistedStop.foreach { stop =>
+        val massTransitStopDirectionValue = stop.validityDirection
+
+        val roadLinkOfMassTransitStop = roadLinkService.getRoadLinkFromVVH(stop.linkId)
```
tvenhola: Could be faster to get all vvh road links by link ids at once: less overhead for HTTP connections, but speedup might be negligible and for a batch job this is good enough.

tvenhola:
Backend should also validate the direction, not just the UI.

UI/src/view/MassTransitStopLayer.js
MAC-CGI: I agree with your sugestion, however it was done this way because the specification says the following -> " If road link traffic direction is "both directions", the default side code value for bus stop can be "towards digitizing". 
Please, note that the backend isn't prepared to work with value 1 (BothDirections).

marquesfm:
Except the commented points, it's all OK for me.

halied:
Ok

UI/src/view/MassTransitStopLayer.js
tvenhola: Okay, let's rename the mapping then. How about renaming the variable as "defaultAssetDirection"?

MAC-CGI: I can't rename it, because it is used in line below (368) to decode the traffic direction.  I can add a comment before for example -> BothDirection varible takes the value 2, to assume the defaultAssetDirection (TowardsDigitizing)

joanacoelho:
Ok

MAC-CGI:
It's Ok for me.

tvenhola:
Ok.

PR #617, opened 2017-02-08
Droth 441 possibility define user role for new user oth

MAC-CGI:


oappicgi:
Looks good, and seems to work as i though it should

joanacoelho:
Looks ok for me

marquesfm:
ok for me.

PR #619, opened 2017-02-16
Drvvh 257 transferring floating road address segments to gap locations

Campanha:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -674,7 +691,7 @@ object RoadAddressDAO {
         case None => ""
       })
       addressPS.setString(11, "-") //TODO: Created by is missing
-    val (p1, p2) = (address.geom.head, address.geom.last)
+    var (p1, p2) = (address.geom.head, address.geom.last)
```
Campanha: No p1, p2 reassignment, so it can be val

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -674,7 +691,7 @@ object RoadAddressDAO {
         case None => ""
       })
       addressPS.setString(11, "-") //TODO: Created by is missing
```
Campanha: Also should add some optional "created_by" to RoadAddress object

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -89,32 +95,79 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
...
+      val roadAddressData = test.roadAddress
+      val sourceIds = test.sourceIds
+      val targetIds = test.targetIds
+      val roadAddresses = roadAddressData.map{ ra =>
+        RoadAddress(ra.id, ra.roadNumber, ra.roadPartNumber, Track.apply(ra.trackCode), Discontinuity.apply(ra.discontinuity), ra.startAddressM, ra.endAddressM,
+        Some(DateTime.now()), None, Option(ra.modifiedBy), ra.linkId, ra.startMValue, ra.endMValue, SideCode.apply(ra.sideCode), ra.calibrationPoints, false, ra.points)
+      }
+
+      roadAddressService.transferFloatingToGap(sourceIds, targetIds, roadAddresses)
+    } catch {
```
pedrosag: Is this try catch mandatory, as in, if this method could throw an exception it should be fixed and patched.

PR #620, opened 2017-02-17
Viite 233 leave out the municipalities of ahvenanmaa from the missing road address script

hestrompa:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/masstransitstop/oracle/queries.scala
```diff
@@ -237,6 +237,14 @@ object Queries {
     """.as[Int].list
   }
 
+  def getMunicipalitiesWithoutAhvenanmaa: Seq[Int] = {
+    var municipalitiesOfAhvenanmaa = List(35, 43, 60, 62, 65, 76, 170, 295, 318, 417, 438, 478, 736, 766, 771, 941)
```
tvenhola: Ahvenanmaa road maintainer id is 0 -> maybe use road maintainer id instead? (if municipalities change the maintainer id should remain)

hestrompa: Ok, I will make the change. Thank you

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/masstransitstop/oracle/queries.scala
```diff
@@ -237,6 +237,13 @@ object Queries {
     """.as[Int].list
   }
 
+  def getMunicipalitiesWithoutAhvenanmaa: Seq[Int] = {
+    var AhvenanmaaRoadMaintainerId = 0
```
Campanha: var -> val and its fine to me

pedrosag: I'm going even deepely than that, if the whole purpose of the variable AhvenanmaaRoadMaintainerId is to be referenced in the SQL with _"#$AhvenanmaaRoadMaintainerId"_ and never to be updated then why isn't this scraped and the SQL switched to _"select id from municipality where **ROAD_MAINTAINER_ID != 0**"_

hestrompa: That was just to identify that the id of Ahvenanmaa road maintainer is 0. But ok, I can remove it

hestrompa: Done!

PR #625, opened 2017-03-03
VIITE-284 User management streamlining

hestrompa:
VIITE-284 User management streamlining
VIITE-282 Hide buttons if user does not have editing permission


JaniKoivisto:
Ok for me

PR #627, opened 2017-03-06
 Update batch process and UI for counting amount of lanes

MAC-CGI:


digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/NumericalLimitFiller.scala
```diff
@@ -162,6 +162,160 @@ object NumericalLimitFiller {
     (segments ++ generated, changeSet)
   }
 
+  private def combine(roadLink: RoadLink, segments: Seq[PersistedLinearAsset], changeSet: ChangeSet): (Seq[PersistedLinearAsset], ChangeSet) = {
```
halied: This seems to be the same code that speed limits use. Is is possible to put it somewhere where both could use it, so there wouldn't be duplicate code? Or are there many differences?

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/NumericalLimitFiller.scala
```diff
@@ -162,6 +162,171 @@ object NumericalLimitFiller {
...
+          sp.value == current.value)
+        if (next.nonEmpty) {
+          return extendOrDivide(Seq(current.copy(endM = next.get.endM)) ++ rest.filterNot(sp => sp.equals(next.get)), segments)
+        }
+      }
+      (segments.copy(sideCode = current.sideCode.value, startMeasure = current.startM, endMeasure = current.endM), rest)
+    }
+
+    /**
+      * Creates speed limits from orphaned segments (segments originating from a speed limit but no longer connected
```
marquesfm: Maybe was copied from the SpeedLimits classes, just modify the comment and it's fine.

marquesfm:
Seems OK to me, except my comment.

halied:
Ok

PR #628, opened 2017-03-06
Viite 269 drawing objects outside boundingbox after siirra

Campanha:


jorgebfrr:
Is better to fix the indentation first. The rest its fine.

PR #629, opened 2017-03-07
DROTH-494 New user role "huoltoteiden ylläpitäjä" with operation area as polygon shape

oappicgi:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/LinearAssetService.scala
```diff
@@ -75,6 +76,24 @@ trait LinearAssetOperations {
...
+        polygonTool.geometryInterceptorToBoundingBox(
+          polygonTool.getAreaGeometry(serviceArea),bounds))
+    for (polygonQueryString<-polygonStringList)
+      {
+        val (roadLinks, change) = roadLinkService.getRoadLinksAndChangesFromVVHWithPolygon(polygonQueryString)
+       roadLinkS++=roadLinks
+      changeS++=change
+      }
+    val linearAssets = getByRoadLinks(typeId,  roadLinkS, changeS)
+    LinearAssetPartitioner.partition(linearAssets, roadLinkS.groupBy(_.linkId).mapValues(_.head))
```
leirasf: I belive we should avoid using var so i belive we can have something like:
```
val vvhRoadlinksAndChanges = polygonStringList.map(roadLinkService.getRoadLinksAndChangesFromVVHWithPolygon)
val roadlinks = vvhRoadlinksAndChanges.map(_._1)
val linearAssets = getByRoadLinks(typeId,  roadlinks, vvhRoadlinksAndChanges.map(_._2))
LinearAssetPartitioner.partition(linearAssets, roadlinks.groupBy(_.linkId).mapValues(_.head))
```

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/LinearAssetService.scala
```diff
@@ -75,6 +76,24 @@ trait LinearAssetOperations {
...
+    val polygonTool = new PolygonTools()
+    var roadLinkS= ListBuffer.empty[RoadLink]
+    var changeS= ListBuffer.empty[ChangeInfo]
+    val polygonStringList =      polygonTool.stringifyGeometryForVVHClient(
+        polygonTool.geometryInterceptorToBoundingBox(
+          polygonTool.getAreaGeometry(serviceArea),bounds))
+    for (polygonQueryString<-polygonStringList)
+      {
+        val (roadLinks, change) = roadLinkService.getRoadLinksAndChangesFromVVHWithPolygon(polygonQueryString)
+       roadLinkS++=roadLinks
```
leirasf: Is not possible to send more than one ring to VVH query, so like this we don't need to do more that one request?

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/util/PolygonTools.scala
```diff
@@ -0,0 +1,91 @@
...
+  def geometryInterceptorToBoundingBox(geometry: Geometry, boundingBox: BoundingRectangle): Seq[Polygon] = {
+    var listGeom = ListBuffer.empty[Polygon]
+    val leftBottomP = boundingBox.leftBottom
+    val rightTopP = boundingBox.rightTop
+    val leftTopP = Point(leftBottomP.x, rightTopP.y)
+    val rightBottom = Point(rightTopP.x, leftBottomP.y)
+    val BoundingBoxAsPoly = geomBuilder.polygon(leftTopP.x, leftTopP.y, rightTopP.x, rightTopP.y, rightBottom.x, rightBottom.y, leftBottomP.x, leftBottomP.y)
+    val intersectionGeometry = geometry.intersection(BoundingBoxAsPoly)
+    if (intersectionGeometry.getGeometryType.toLowerCase.startsWith("polygon")) {
+      listGeom += intersectionGeometry.asInstanceOf[Polygon]
```
leirasf: It seems to me that the `listGeom` is not used. Maybe we just need to have something like Seq(intersectionGeometry.asInstanceOf[Polygon]).
At the same time we get ride of a var :)

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/LinearAssetService.scala
oappicgi: All arcserver documentation examples showed only one polygon with rings per request. I  tried to send request with different variations  to VVH, but I received either error or it processed request extremely long and returned error message saying request could not be processed.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/util/PolygonTools.scala
oappicgi: yea, at first i planned on making multipolygon-> polygon sequence conversion in that method but now that it is in different method we don't need buffered list there anymore

joanacoelho:
Ok for me

halied:
Ok

UI/src/view/navigationpanel/LinearAssetBox.js
```diff
@@ -31,7 +31,11 @@
     var bindExternalEventHandlers = function() {
       eventbus.on('roles:fetched', function(roles) {
         userRoles = roles;
-        if (_.contains(roles, 'operator') || _.contains(roles, 'premium')) {
+        if (_.contains(roles, 'operator') || (_.contains(roles, 'premium') && !elements.expanded.hasClass('panel maintenanceRoad'))) {
```
marquesfm: we Can join this IF clause with ELSE IF clause below?
I say that, because the IF clause and the ELSE IF clause do the same code.

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -848,6 +883,7 @@ Returns empty result as Json message, not as page not found
   }
 
   private def validateUserMunicipalityAccess(user: User)(municipality: Int): Unit = {
+    if (!user.isServiceRoadMaintainer())
```
marquesfm: This IF doesn't do nothing.
The idea was to put the halt(Unauthorized("ServiceRoad user is only authorized to alter serviceroad assets")) like was placed in the others IF's?

marquesfm:
With the exception of my comments, it's ok.

PR #638, opened 2017-03-15
VIITE-249 Reservation of road parts for road address projects

oappicgi:
Doesn't allow user to reserve links already reserved by other project


JaniKoivisto:
OK

PR #649, opened 2017-03-28
Droth 546 automatically generated info to kalpa api

oskarmalmstrom:


digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/IntegrationApi.scala
```diff
@@ -240,6 +246,33 @@ class IntegrationApi(val massTransitStopService: MassTransitStopService) extends
...
+
+    val nonFixedValues = List("split_speedlimit_.*", "batch_process_.*")
+
+    val nonFixedModified = nonFixedValues match {
+      case value if value.exists(nfv => nfv.r.findFirstIn(modifiedBy.getOrElse("")).equals(modifiedBy)) =>
+        modifiedBy
+      case _ => None
+    }
+    val nonFixedCreated = nonFixedValues match {
+      case value if value.exists(nfv => nfv.r.findFirstIn(modifiedBy.getOrElse("")).equals(modifiedBy)) =>
```
joanacoelho: shouldn't be used the createdBy instead of modifiedBy?  And if so, you can do like this probably : 

```
def lastModifiedBy(...) = {
	val nonFixedModified = obtainFixedValues(modifiedBy)
	val nonFixedCreated = obtainFixedValues(createdBy)
  (...)
}

def obtainFixedValues( str : Option[String] ) {
    val nonFixedValues = List("split_speedlimit_.*", "batch_process_.*")
     nonFixedValues match {
	case value if value.exists(nfv => nfv.r.findFirstIn(str.getOrElse("")).equals(str)) => str
	case _ => None
	}
  }
```

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/IntegrationApi.scala
```diff
@@ -240,6 +246,33 @@ class IntegrationApi(val massTransitStopService: MassTransitStopService) extends
...
+        createdBy
+      case _ => None
+    }
+
+    val autoGeneratedValues = List("dr1conversion", "dr1_conversion", nonFixedModified.getOrElse(""), nonFixedCreated.getOrElse(""), "automatic_correction", "excel_data_migration", "automatic_generation", "vvh_generated")
+
+    modifiedBy match{
+      case None => "generatedValue" -> false
+      case Some(value) if(value != null) => if (autoGeneratedValues.exists(agv => agv.equals(modifiedBy.getOrElse(""))))return "generatedValue" -> true else return "generatedValue" -> false
+    }
```
joanacoelho: Can you try like this: 

```
modifiedBy match{

   case Some(value) if value != null && autoGeneratedValues.exists(agv => agv.equals(modifiedBy.getOrElse(""))) => 

     "generatedValue" -> true 

   case None => "generatedValue" -> false
```

I believe that with this you can remove the `else return "generatedValue" -> false`

joanacoelho:
Everything else seems ok

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/IntegrationApi.scala
```diff
@@ -240,6 +246,33 @@ class IntegrationApi(val massTransitStopService: MassTransitStopService) extends
         .getOrElse("")
   }
 
+  def lastModifiedBy(createdBy: Option[String], modifiedBy: Option[String]): (String, Boolean) = {
+
```
leirasf: Instead of having the regex validation you cold have something like that

```
def nonFixedUsers(userOption: Option[String])={
      val nonFixedValues = List("split_speedlimit_", "batch_process_")
      userOption match {
        case Some(user) if(nonFixedValues.exists(v => user.startsWith(v))) =>
          user
        case _ =>
          ""
      }
    }
```

And then you can using direclly in the list 

```
val autoGeneratedValues = List("dr1conversion", "dr1_conversion", nonFixedUsers(modifiedBy), nonFixedUsers(createdBy), "automatic_correction", "excel_data_migration", "automatic_generation", "vvh_generated")
```

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/IntegrationApi.scala
```diff
@@ -240,6 +246,33 @@ class IntegrationApi(val massTransitStopService: MassTransitStopService) extends
...
+    modifiedBy match{
+      case None => "generatedValue" -> false
+      case Some(value) if(value != null) => if (autoGeneratedValues.exists(agv => agv.equals(modifiedBy.getOrElse(""))))return "generatedValue" -> true else return "generatedValue" -> false
+    }
+    createdBy match{
+      case None => "generatedValue" -> false
+      case Some(value) if(modifiedBy.getOrElse("") != null || modifiedBy.getOrElse("") == null) => "generatedValue" -> autoGeneratedValues.exists(agv => agv.equals(value))
+    }
+  }
```
leirasf: You can have something like:

 ```
modifiedBy match{
      case None => "generatedValue" -> false
      case Some(value) if(value != null) => 
        "generatedValue" -> autoGeneratedValues.exists(agv => agv.equals(value))
      case _ =>
        createdBy match{
          case Some(value) =>
            "generatedValue" -> autoGeneratedValues.exists(agv => agv.equals(value))
          case _ => "generatedValue" -> false
        }
    }
```

MAC-CGI:
For me its OK.

marquesfm:
ok to me.

PR #655, opened 2017-04-04
Viite-317 - Added startDate option to road address creation && Viite-314 - Removal of Sulje-button from UI

pedrosag:


pedrosag:
Please do keep in mind that this also gets the fix for the [task 314](https://extranet.liikennevirasto.fi/jira/browse/VIITE-314) .

PR #656, opened 2017-04-04
DROTH-490 Point-like asset bearing value needs to be adjusted in OTH

MAC-CGI:
DROTH-490 Point-like asset bearing value needs to be adjusted in OTH


halied:
Ok

leirasf:
It looks good to me

PR #657, opened 2017-04-05
VIITE-318_CREATED_BY_gets_the_username_of_the_user_who_moved_the_floating

jorgebfrr:
Add current username in the database, when clicked on Tallena button.


jorgebfrr:
@hestrompa i have already made the changes, is possible to validate again?
@Campanha too.

PR #660, opened 2017-04-07
DROTH-552 csv import is not working properly, no data can be updated …

marquesfm:
…to TR or OTH

- Create a validation to verify if the property "tietojen_yllapitaja" it's coming with empty values, in that case don't update the Property and keep with the old value;
- Create a new unit test for validate property yllapitajan_koodi , needs to be empty because property tietojen_yllapitaja is empty in the properties received.


marquesfm:
Ok for me....

joanacoelho:
Ok for me

PR #668, opened 2017-04-13
Viite 304 creation of process to update road address geometry

hestrompa:


digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/util/DataFixture.scala
```diff
@@ -160,9 +168,12 @@ object DataFixture {
         recalculate()
       case Some ("update_missing") =>
         updateMissingRoadAddresses()
-      case Some("fuse_multi_segment_road_addresses") => {
+      case Some("fuse_multi_segment_road_addresses") =>
         combineMultipleSegmentsOnLinks()
-      }
+      case Some("update_road_addresses_geometry_staging") =>
```
tvenhola: Why separate jobs and not just using the properties to figure out the environment?

hestrompa: I will rename both processes to "update_road_addresses_geometry_no_complementary" and "update_road_addresses_geometry", to make processes more understandable

PR #673, opened 2017-04-18
DROTH-568 If a bus stop (maintainer ELY) status is moved from anything to virtual (type = 5) stop needs to be removed from Tierekisteri

marquesfm:
- Create method to verified if BusStop was modified to Virtual;
- Join the previous validation created to remove the LiVi_ID in OTH in case bus stop (maintainer ELY) status was moved from anything to virtual;
- Add the Remove Operation type in the Operation match.

### DIFF
--- a/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/MassTransitStopService.scala
+++ b/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/MassTransitStopService.scala
@@ -423,12 +423,13 @@ trait MassTransitStopService extends PointAssetOperations {
 
       val id = asset.id
       massTransitStopDao.updateAssetLastModified(id, username)
+      val isVirtualBusStop = MassTransitStopOperations.isVirtualBusStop(properties)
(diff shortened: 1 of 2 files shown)


halied:
I get this error when I try starting server in this branch:
[info] Compiling 1 Scala source to C:\Digiroad\uus\digiroad2\digiroad2-api-viite\target\scala-2.11\classes...
[error] C:\Digiroad\uus\digiroad2\digiroad2-api-viite\src\main\scala\fi\liikennevirasto\digiroad2\ViiteApi.scala:157: value size is not a member of (Option[fi.liikennevirasto.viite.dao.CalibrationPoint], Option[fi.liikennevirasto.viite.dao.CalibrationPoint])
[error]       val pointsToCalibration = ra.calibrationPoints.size match {

marquesfm:
@halied I will talked with Viite Team about that.
It's a problem with their API.

halied:
What do you mean with "any more problems in the code"?

joanacoelho:
Ok to me

marquesfm:
@halied I mean, if it's all ok or if you suggest some upgrades in the code?

halied:
Ok

PR #675, opened 2017-04-19
Viite 304 road address segment is moved over 1 meter

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -529,6 +529,16 @@ object RoadAddressDAO {
       """.as[Long].list
   }
 
+  def getValidRoadNumbersWithFilter = {
```
oappicgi: I did not understand/remember why we couldn't use previous method (one with out filter)

oappicgi:
looks good. if "getValidRoadNumbersWithFilter" is only meant to be used for testing environment it would be nice to see that in name... for example "getValidRoadNumbersWithFilterToTestEnv"

tvenhola:
New pull request after logical problems are handled the way the new specification describes them.

PR #683, opened 2017-04-26
DROTH-608: Road link modified date comparison fix

halied:
Show the latest of road link modification dates (property modification date or vvh modification date) in UI form


leirasf:
Looks good to me

PR #685, opened 2017-05-02
VIITE-367 Calibration point after pressing siirrä

hestrompa:
Added calibration points to sirraa temporary layer


hestrompa:
@oappicgi , I think the problem you reported is now solved. Can you validate again please?

PR #691, opened 2017-05-04
Viite 358 - floating road address are still drawn on the map

pedrosag:
I used this task to reflect the following 2 changes:

Change 1 of 2 - Eliminating the Unknown Markers on the map after Clicking "Siirrä".

Change  2 of 2 - Filtering the results of the backend call in order to omit the features that are already "in simulation".
                    Also needed to make sure that when we are cancelling the simulation, we need to get the roadlinks again, because the roadlayer is missing the ones that were omitted.


viite-UI/src/view/LinkPropertyLayer.js
```diff
@@ -946,30 +967,78 @@
       selectedLinkProperty.setFloatingRoadMarker(floatingRoadMarker);
     };
 
+    var cleanUnknowsAfterDefloat = function(){
```
Campanha: did you mean cleanUnknow(n)sAfterDefloat? How could you miss that ? :D

PR #692, opened 2017-05-05
Droth 588 point assets snapping to link

tvenhola:
Make point assets snap to road link if geometry is adjusted a bit but not enough to make it float


leirasf:
Looks good to me

marquesfm:
Looks Ok to me

joanacoelho:
Ok for me

PR #694, opened 2017-05-08
Viite fixes road address project

JoaoSVerissimo:
Fixes issues:
VIITE-399
VIITE-384
VIITE-388
VIITE-397


viite-UI/src/view/RoadAddressProjectForm.js
```diff
@@ -157,16 +157,20 @@
...
         applicationModel.setOpenProject(true);
+        projectCollection.clearRoadAddressProjects();
       });
 
       eventbus.on('roadAddress:openProject', function(result) {
         currentProject = result.projects;
+        projectCollection.clearRoadAddressProjects();
         var text = '';
-        _.each(result.projectLinks, function(line){
+        _.each(result.projectLinks, function(line){  //TODO later list of already saved roadlinks has to be saved in  roadaddressprojectcollection.currentRoadSegmentList for reserve button to function properly now saved links are cleared when newones are reserved
```
pedrosag: Sorry to bother @oappicgi, but is this TODO, to be done in a separate task?

I ask this since it was this [commit](https://github.com/finnishtransportagency/digiroad2/pull/694/commits/988477a407f837635028675bdf3d851669c14a9e) that introduced the TODO.

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -351,7 +351,7 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
     Map("roadNumber" -> roadAddressProjectFormLine.roadNumber,
       "roadPartNumber" -> roadAddressProjectFormLine.roadNumber,
```
Campanha: those seems not to be used yet but will be used in the future?

viite-UI/src/view/RoadAddressProjectForm.js
oappicgi: It wasn't included in this task. PO said there will be task  to manage old projects later, so it can be either removed or left there as reminder. To my understanding it will be implemented after we get submitting projects to TR working.

pedrosag: It's fine, better to leave that as a reminder then ;)

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
oappicgi: good catch. For some reason idea did not mark that as not used for me. It also has       "roadPartId" -> roadAddressProjectFormLine.roadNumber,  which doesn't seem right when roadpart and roadpartid are different things  in reservedRoadPartToApi method.  Probably better to remove it and if really needed, rewrite it.

PR #697, opened 2017-05-10
Viite 406 transfer fails in certain conditions

tvenhola:
Road address transfers failed under more complex requests and couldn't produce proper results.

This pull request should fix all known issues with roads 847, 848, 15580.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/DefloatMapper.scala
```diff
@@ -0,0 +1,312 @@
...
+    val startCPAddr = seq.find(_.startCalibrationPoint.nonEmpty)
+    val endCPAddr = seq.find(_.endCalibrationPoint.nonEmpty)
+    if (startCPAddr.nonEmpty) {
+      val (addr, cp) = (startCPAddr.get, startCPAddr.get.startCalibrationPoint.get)
+      if (addr.startAddrMValue != cp.addressMValue)
+        throw new IllegalArgumentException(s"Start calibration point value mismatch in $cp")
+      if (seq.exists(_.startAddrMValue < cp.addressMValue))
+        throw new IllegalArgumentException("Start calibration point not in the first link of source")
+      if (addr.sideCode == SideCode.TowardsDigitizing && Math.abs(cp.segmentMValue) > 0.0 ||
+        addr.sideCode == SideCode.AgainstDigitizing && Math.abs(cp.segmentMValue - addr.endMValue) > 0.001)
```
Campanha: should we use MaxAllowedMValueError instead of 0.001? Still i dunno if we will be use this const again elsewhere

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -685,165 +634,25 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
-    val maxEndAddressM = orderedSources.last.endAddressM
-
-    val adjustedSegments = adjustTopology.foldLeft(orderedSources) { (previousSources, operation) => operation(targetsGeomLength, previousSources) }
-
-    val maxEndMValue = allLinks.flatMap(_.endCalibrationPoint) match {
-      case Nil => adjustedSegments.map(_.endMValue).max
-      case _ => getMValues(Option(adjustedSegments.flatMap(_.endCalibrationPoint).map(_.segmentMValue).max), Option(allLinks.map(_.endMValue).max))
+  def transferRoadAddress(sources: Seq[RoadAddressLink], targets: Seq[RoadAddressLink], user: User): Seq[RoadAddress] = {
+    val mapping = DefloatMapper.createAddressMap(sources, targets)
+    if (mapping.exists(DefloatMapper.invalidMapping)) {
```
oappicgi: naming logic seems odd

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/DefloatMapper.scala
tvenhola: We should, I believe

Campanha: ok, better use constant then

PR #700, opened 2017-05-12
Viite 414 disable navigation panel buttons when editing

JoaoSVerissimo:


tvenhola:
Small change, tested on UI.

PR #707, opened 2017-05-17
VIITE 413 - T intersection doesn't give an error

JoaoSVerissimo:


tvenhola:
DefloatMapperSpec fails 3/5 tests. Must pass the valid geometry tests before merging.

PR #715, opened 2017-05-22
Viite 413 t intersection doesn't give an error

tvenhola:
Fixed matching rules to allow non-intersecting but non-connected geometries as valid targets.


tvenhola:
Cleaned up a bit. Merging.

PR #719, opened 2017-05-25
Viite 375 save project link status

JaniKoivisto:


viite-UI/src/model/RoadAddressProjectCollection.js
```diff
@@ -135,22 +139,29 @@
...
       var data = {'linkIds': linkIds, 'projectId': projectId, 'newStatus': STATUS_TERMINATED};
 
+      console.log(data);
+      console.log(linkIds);
       if(!_.isEmpty(linkIds) && typeof projectId !== 'undefined' && projectId !== 0){
         backend.updateProjectLinks(data, function(errorObject) {
           if (errorObject.status == INTERNAL_SERVER_ERROR_500 || errorObject.status == BAD_REQUEST_400) {
             eventbus.trigger('roadAddress:projectLinksUpdateFailed', errorObject.status);
+          } else {
+            console.log("TODO Success, refresh now");
```
Campanha: i think we already have the refresh in success case. "roadAddress:projectLinksUpdated" is the event that will set the current selected roads as dirty and redraw them. Can someone check this?

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -197,6 +197,17 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
     }
   }
 
+  post("/roadlinks/roadaddress/project/sendToTR"){
+   try{
+     val projectID = parsedBody.extract[projectID]
+    //TODO call getRoadAddressChangesAndSendToTR(projectID). if success change projectstatus to "sent to TR"
```
pedrosag: Is this to be done in a separated task?

viite-UI/src/view/RoadAddressProjectEditForm.js
```diff
@@ -0,0 +1,212 @@
...
+    var selectedProjectLink = false;
+    var staticField = function(labelText, dataField) {
+      var field;
+      field = '<div class="form-group">' +
+        '<p class="form-control-static asset-log-info">' + labelText + ' : ' + dataField + '</p>' +
+        '</div>';
+      return field;
+    };
+    var actionSelectedField = function() {
+      //TODO: cancel and save buttons Viite-374
```
pedrosag: Viite-374 is already on master, is it ok to delete this?

Campanha: i think the cancel is missing but save buttons are ok

hestrompa: The save button is working, but the cancel still have some issues to solve on VIITE-377 (Form after user has clicked "Peruuta"(Cancel))

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
oappicgi: I think it would be easier to add this on Viite-380 task, when we have both 375 and 64 on master.

viite-UI/src/model/RoadAddressProjectCollection.js
tvenhola: Dirty data means that the data is modified in the UI but not yet saved to the database. After successful save we should reload the road links from the backend to bring the latest view.

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
tvenhola: This is a 380 (/ 379) task

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -31,7 +31,7 @@ case class NewAddressDataExtracted(sourceIds: Set[Long], targetIds: Set[Long], r
 case class RoadAddressProjectExtractor(id: Long, status: Long, name: String, startDate: String, additionalInfo: String,roadPartList: List[ReservedRoadPart])
 
 case class RoadAddressProjectLinkUpdate(linkIds: Seq[Long], projectId: Long, newStatus: Int)
-
+case class projectID(projectID:Long)
```
tvenhola: Do we really need a case class to wrap a Long?

viite-UI/src/view/ProjectLinkLayer.js
```diff
@@ -92,6 +92,7 @@
           return !_.isUndefined(selectionTarget.projectLinkData);
```
Campanha: Its not visible since it comes from viite-374 but in line 90 we have one "// TODO: 374 validate selection" in this selectedSingleClick control. What kind of validation should it be?

viite-UI/src/model/RoadAddressProjectCollection.js
hestrompa: The code is never entering in the else clause, because this is only a failure event. When we have success on the saving process, we are calling the roadAddress:projectLinksUpdated event, and the roads are updated there.

hestrompa: Calling the fetch now after saving on database, in order to refresh the view with the saved data. I'll commit the changes

viite-UI/src/view/ProjectLinkLayer.js
tvenhola: That is related to 373, "User may only apply selection to Segment in progress"

Although later it should allow selecting non-addressed road links as well.

PR #723, opened 2017-05-26
Viite 304 road address segment is moved over 1 meter

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -343,15 +343,21 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
+  def checkRoadAddressFloatingWithoutTX(ids: Set[Long], float: Boolean = false): Unit = {
     val addresses = RoadAddressDAO.queryById(ids)
     val linkIdMap = addresses.groupBy(_.linkId).mapValues(_.map(_.id))
     val roadLinks = roadLinkService.getCurrentAndComplementaryVVHRoadLinks(linkIdMap.keySet)
     addresses.foreach { address =>
       val roadLink = roadLinks.find(_.linkId == address.linkId)
       val addressGeometry = roadLink.map(rl =>
         GeometryUtils.truncateGeometry3D(rl.geometry, address.startMValue, address.endMValue))
-      if (roadLink.isEmpty || addressGeometry.isEmpty || GeometryUtils.geometryLength(addressGeometry.get) == 0.0) {
+      if(float && !(roadLink.isEmpty || addressGeometry.isEmpty || GeometryUtils.geometryLength(addressGeometry.get) == 0.0)){
```
tvenhola: We should document a bit what's happening here - it's not clear at a glance

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -557,6 +564,16 @@ object RoadAddressDAO {
       """.as[Long].list
   }
 
+  def getValidRoadNumbersWithFilterToTestAndDevEnv = {
+    sql"""
+       select distinct road_number
+              from road_address ra
+              where ra.floating = '0' AND (end_date < sysdate OR end_date IS NULL) AND
```
tvenhola: Should check only valid rows (valid_to > sysdate OR valid_to is NULL)

PR #724, opened 2017-05-29
DROTH-667 authorization validation, new header to trclient

oskarmalmstrom:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/TierekisteriClient.scala
```diff
@@ -186,7 +186,8 @@ trait TierekisteriClient{
 
   protected def request[T](url: String): Either[T, TierekisteriError] = {
     val request = new HttpGet(url)
-    request.addHeader("X-OTH-Authorization", "Basic " + auth.getAuthInBase64)
+    request.addHeader("X-OTH-Authorization", "Basic " + auth.getOldAuthInBase64)
+    request.addHeader("X-Authorization", "Basic " + auth.getAuthInBase64)
```
leirasf: We can have a method to set the headers with the authentication so we don't need to have it duplicated.

joanacoelho:
ok

PR #727, opened 2017-05-30
Droth 663 when a linear asset is expired mvalues are updated

MAC-CGI:
Droth-663 when a linear asset is expired mvalues are updated


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/oracle/OracleLinearAssetDao.scala
```diff
@@ -1115,6 +1115,12 @@ class OracleLinearAssetDao(val vvhClient: VVHClient) {
     }
 
     insertProhibitionValue(id, value)
+    (startMeasure, endMeasure) match {
+      case (None, None) => None
```
joanacoelho: try this : 
```
case  (Some(start), Some(end)) => updateMValues(id, (start, end))
case _ => None
```

halied:
Ok

joanacoelho:
Ok for me

joanacoelho:
ok for me

marquesfm:
ok for me

PR #731, opened 2017-05-31
DROTH-671 (US 661) Create the new Integration API entry point

joanacoelho:


joanacoelho:
Ok

marquesfm:
Ok

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/IntegrationApi.scala
```diff
@@ -353,6 +369,31 @@ class IntegrationApi(val massTransitStopService: MassTransitStopService) extends
...
+    } else if (createdDateTime.exists(_.isAfter(since))) {
+      "Add"
+    } else {
+      "Modify"
+    }
+  }
+
+  get("/changes/:assetType") {
+    contentType = formats("json")
+     val since = DateTime.parse(params("since"))
```
leirasf: We should have something like params.getOrElse("since", halt(BadRequest("Missing mandatory 'since' parameter")))

halied:
Ok

PR #732, opened 2017-06-01
Viite 422 move project links to road address table

pedrosag:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -431,4 +434,26 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
       case _=> None
     }
   }
+
+  def updateRoadAddressWithProject(newState: ProjectState, projectID: Long) ={
+    if((newState.value != ErroredInTR.value || newState.value != Unknown.value) && newState.value == Saved2TR.value){
```
tvenhola: doesn't this evaluate to newState == Saved2TR?

PR #735, opened 2017-06-01
Viite 423 424 project dates error handling

hestrompa:


oappicgi:
seems bit wasteful to loop list twice in projDateValidation, but i guess it was earlier like that

PR #741, opened 2017-06-06
VIITE-405 reset roadcollection before fetching new road links 

tvenhola:
Force refreshing to get new road link address values for the whole road.


tvenhola:
Merging one line fix.

PR #745, opened 2017-06-07
Viite 377 form after user has clicked peruuta(cancel)

JoaoSVerissimo:


oappicgi:
getting  1 failing

  1) when click on the Tieosoiteprojektit button when clicking Peruuta button Check if it change to the road form:
     AssertionError: Form didn't open.

i will do fixture reset if that resolves it

PR #746, opened 2017-06-07
DROTH-660_linelike_assets_to_complimentary_roadlinks

oskarmalmstrom:


UI/src/asset-type-layer-specifications.js
```diff
@@ -40,6 +40,7 @@
       className: 'total-weight-limit',
       unit: 'kg',
       isSeparable: false,
+      complementaryPossible: true,
```
leirasf: I think allowComplementaryLinks it's a better name for that configuration.

joanacoelho:
ok

PR #747, opened 2017-06-07
join Groupby by Anomaly where 1 or 2

Campanha:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/model/RoadAddressLinkPartitioner.scala
```diff
@@ -8,7 +8,7 @@ object RoadAddressLinkPartitioner extends GraphPartitioner {
 
   def partition[T <: RoadAddressLinkLike](links: Seq[T]): Seq[Seq[T]] = {
     val linkGroups = links.groupBy { link => (
-      link.anomaly.value, link.roadNumber, link.roadPartNumber, link.trackCode,
+      link.anomaly.value.equals(Anomaly.None), link.anomaly.value.equals(Anomaly.Illogical), link.roadNumber, link.roadPartNumber, link.trackCode,
```
tvenhola: As long as we're not using Illogical (and we're not sure if we ever will) we might just remove that comparison and be happy to check if there is any anomaly or none.

tvenhola: Also, comparison of unrelated object types: link.anomaly.value <=> Anomaly

tvenhola:
This is ok

PR #748, opened 2017-06-08
Viite 492: hower info popup for roadaddress in project link action selection mode

oappicgi:
hower

### DIFF
--- a/viite-UI/src/view/ProjectLinkLayer.js
+++ b/viite-UI/src/view/ProjectLinkLayer.js
@@ -222,48 +222,50 @@
 
     map.addOverlay(overlay);
 
-    //Listen pointermove and get pixel for displaying roadaddress feature info
+    //Listen pointerMove and get pixel for displaying roadAddress feature info
     eventbus.on('map:mouseMoved', function (event, pixel) {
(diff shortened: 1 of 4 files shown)


viite-UI/src/view/ProjectLinkLayer.js
```diff
@@ -212,6 +212,65 @@
     //This will control the double click zoom when there is no selection that activates
     map.on('dblclick', zoomDoubleClickListener);
 
+    var infoContainer = document.getElementById('popup');
+    var infoContent = document.getElementById('popup-content');
+
+    var overlay = new ol.Overlay(({
+      element: infoContainer
+      //autoPan: false,
```
pedrosag: Is this one needed?

Campanha:
this is a nice feature!

viite-UI/src/view/ProjectLinkLayer.js
oappicgi: yea, we tested that with Jani and forgot it there. if enabled popup would in cases where it is near corners of the map  push map in direction it appears.

tvenhola:
Looks good

PR #751, opened 2017-06-08
VIITE-467 Error in end_date in ROAD_ADDRESS table.

hestrompa:
I made some changes on queryList method, because the case clauses only takes a max of 22 fields, and we need to validate 23 fields. To solve that problem, I created a "implicit val getRoadAddressChangeRow" that handles those 23 values


tvenhola:
This is good [+]

PR #753, opened 2017-06-14
Fix for not parsing actual error message

oappicgi:
When tierekisteri did not accept project, user  always got "Bad request" message to UI after pressing publish. Now we parse error message when we are expecting one (with html error codes 400-499)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ViiteTierekisteriClient.scala
```diff
@@ -182,14 +183,17 @@ object ViiteTierekisteriClient {
...
     request.setEntity(createJsonmessage(trProject))
     val response = client.execute(request)
+    var errorMessage= TRErrorResponse("")
     try {
-    val statusCode = response.getStatusLine.getStatusCode
-    val reason = response.getStatusLine.getReasonPhrase
-    ProjectChangeStatus(trProject.id, statusCode, reason)
+      val statusCode = response.getStatusLine.getStatusCode
+      if (statusCode>=400 && statusCode<500)
+      errorMessage = parse(StreamInput(response.getEntity.getContent)).extract[TRErrorResponse] // would be nice if we didn't need case class for parsing of one attribute
```
Campanha: would it be possible to have something like parse(StreamInput(response.getEntity.getContent)).extractOpt[TRErrorResponse]..getOrElse(TRErrorResponse(""))
So we didnt need to declare errorMessage as var

PR #760, opened 2017-06-19
Viite 493 vvh changetable

pedrosag:
Changed the RoadAddress in order to include the adjusted_timestamp~from LRM_POSITION; ground work for Viite-82.

Added a test for the getChangeInfoFromVVH method on the RoadLinkService.

Added a small helper method that when supplied with both with a sequence of RoadAddresses and ChangeInfos it will return what un-applied changeInfos match said roadAddress.

Added a TODO as a reminder of future changes to the vvhChangeInfo queries.

### DIFF
--- a/digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/RoadLinkServiceSpec.scala
+++ b/digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/RoadLinkServiceSpec.scala
@@ -864,4 +864,26 @@ class RoadLinkServiceSpec extends FunSuite with Matchers with BeforeAndAfter {
     }
   }
 
+  test("verify the output of change info") {
+    val oldLinkId = 1l
+    val newLinkId = 2l
+    val changeInfo = ChangeInfo(Some(oldLinkId), Some(newLinkId), 123l, 5, Some(0), Some(1), Some(0), Some(1), 144000000)
+    val boundingBox = BoundingRectangle(Point(123, 345), Point(567, 678))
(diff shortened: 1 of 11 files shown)


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/RoadLinkServiceSpec.scala
```diff
@@ -864,4 +864,26 @@ class RoadLinkServiceSpec extends FunSuite with Matchers with BeforeAndAfter {
...
+  test("verify the output of change info") {
+    val oldLinkId = 1l
+    val newLinkId = 2l
+    val changeInfo = ChangeInfo(Some(oldLinkId), Some(newLinkId), 123l, 5, Some(0), Some(1), Some(0), Some(1), 144000000)
+    val boundingBox = BoundingRectangle(Point(123, 345), Point(567, 678))
+    val mockVVHClient = MockitoSugar.mock[VVHClient]
+    val mockVVHRoadLinkClient = MockitoSugar.mock[VVHRoadLinkClient]
+    val mockVVHChangeInfoClient = MockitoSugar.mock[VVHChangeInfoClient]
+    val service = new TestService(mockVVHClient)
+    OracleDatabase.withDynTransaction {
```
JoaoSVerissimo: Hi, It's better to start using the overrided withDynTransaction method call.

pedrosag: Maybe but if you look at all the tests in the test suit all of them are not using the overided version.

PR #761, opened 2017-06-19
Viite 430 show direction of road address

Campanha:
Feature of showing digitizing orientation similar to OTH manoeuvres.


viite-UI/src/view/LinkPropertyLayer.js
```diff
@@ -518,6 +526,14 @@
             floatingMarkerLayer.getSource().addFeature(marker);
         });
 
+        var directionRMIds = _.pluck(directionRoadMarker, 'linkId');
```
JoaoSVerissimo: Maybe we need stop using the pluck and use the map instead, because, pluck was removed from lodash 4.

PR #767, opened 2017-06-19
Droth 685 labeling of linear assets numerical values

joanacoelho:


halied:
Ok

joanacoelho:
ok

UI/src/view/LinearAssetLabel.js
```diff
@@ -4,22 +4,47 @@
...
+              }))
+            });
+          }else if (valueLength > 4 && valueLength < 7) {
+            return new ol.style.Style({
+              image: new ol.style.Icon(({
+                src: 'images/linearLabel_background_large.png'
+              }))
+            });
+          }
+          return new ol.style.Style({
```
MAC-CGI: The code before could be simplified with something like that
            switch (true){
                case (valueLength > 6 || value < 0):
                    src_image = 'images/warningLabel.png';
                    break;
                case (valueLength > 4 && valueLength < 7):
                    src_image = 'images/linearLabel_background_large.png';
                    break;
                default:
                    src_image = 'images/linearLabel_background.png';
                    break;
            }

PR #771, opened 2017-06-22
Viite 519 add new columns to project change table

JaniKoivisto:
Add three new columns to table and some layout fix


viite-UI/src/less/site/project-change-table.less
```diff
@@ -42,22 +42,22 @@
...
   font-weight: bold;
   height: 28px;
   font-size: 12px;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
 }
 .project-change-table-dimension-first {
-  width: 133px;
+  width: 128px;
```
tvenhola: Prefer "em" as the measurement on text containers. For pictures, margins etc. px is good but font size should define the table size.

viite-UI/src/view/ProjectChangeTable.js
```diff
@@ -23,27 +23,30 @@
       '<div id ="change-table-borders-target"></div>');
     changeTable.append('<div class="change-table-sections">' +
       '<label class="change-table-heading-label" style="width: 117px">Ilmoitus</label>' +
-      '<label class="change-table-heading-label" style="width: 379px">Nykyosoite</label>' +
+      '<label class="change-table-heading-label" style="width: 478px">Nykyosoite</label>' +
```
tvenhola: use classes instead of html tag specific style attributes

viite-UI/src/less/site/project-change-table.less
```diff
@@ -68,10 +68,14 @@
   overflow-y: auto;
 }
 .project-change-table-data-cell {
-  width: 66px;
   color: #5a5a57;
   height: 28px;
   font-size: 12px;
+  overflow: hidden;
+  width: 54.8px;
```
tvenhola: pixel widths, font sizes. Not good for non-desktop users such as tablets etc.

viite-UI/src/view/ProjectChangeTable.js
JaniKoivisto: Right, will change those, its more clear that way

PR #783, opened 2017-06-27
DROTH-746_Change_infotext_on_bus_stop_direction_creation

oskarmalmstrom:
first checks if road is one directional, then if the stop direction is the same as the road direction

### DIFF
--- a/UI/src/model/SelectedMassTransitStop.js
+++ b/UI/src/model/SelectedMassTransitStop.js
@@ -277,6 +277,21 @@
         }
     };
 
+    var validateDirectionsForCreation = function () {
+      if(roadCollection) {
+        var nearestLine = geometrycalculator.findNearestLine(roadCollection.getRoadsForMassTransitStops(), currentAsset.payload.lon, currentAsset.payload.lat);
+        var directions_decode = {BothDirections: 1, TowardsDigitizing: 2, AgainstDigitizing: 3};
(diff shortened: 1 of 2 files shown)


joanacoelho:
looks good

UI/src/model/SelectedMassTransitStop.js
```diff
@@ -277,6 +277,21 @@
...
+        var stopDirection = get('validityDirection');
+
+        if(roadLinkDirection != 1)
+          return stopDirection != roadLinkDirection;
+        return true;
+      }else{
+        return false;
+      }
+    };
+
```
MAC-CGI: validateDirectionsForSave and validateDirectionsForCreation have a similar code only with a different return.
Could you probably create a function to return the RoadLinkDirection. Something like that:

var getRoadLinkDirection = function(){
      var nearestLine = geometrycalculator.findNearestLine(roadCollection.getRoadsForMassTransitStops(), currentAsset.payload.lon, currentAsset.payload.lat);
      var linkId = nearestLine.linkId;
      if (!currentAsset.linkId)
        currentAsset.linkId = linkId;
      var directions_decode = {BothDirections: 1, TowardsDigitizing: 2, AgainstDigitizing: 3};
      return directions_decode[nearestLine.trafficDirection];
    };

halied:
Ok

marquesfm:
Ok for me

PR #788, opened 2017-06-29
DROTH-728 TR changes: Bus stops with end date should be modifiable in OTH

oskarmalmstrom:
Bus stops are now modifiable(form is read only) if: They have an end date in the future, have no end date
Not modifiable if the end day is older than today.


marquesfm:
Ok for me.

halied:
Ok

joanacoelho:
Looks good

PR #794, opened 2017-07-03
Viite 437 partitioner of project links

Campanha:


oappicgi:
code looks good, i will also run checks and visually test it

PR #798, opened 2017-07-05
DROTH-727 When moving bus stops more 50 mts (in state roads) inventointipäivä (inventory date) is not created for the newly created stop

marquesfm:
- Chnage the value of loppupvm and inventointipvm to moving date when moving bus stops more 50 mts.


joanacoelho:
Looks good

oskarmalmstrom:
ok

PR #805, opened 2017-07-07
Quick fixes board

plourenco90:
VIITE-550 Direction markers are visible in wrong scales when zooming out.
VIITE-552 Direction markers on grey dashed lines are pink, not grey and on turquoise dashed lines they are grey, not turquoise.
VIITE-556 Show the direction markers after "Siirrä" when transferring floatings.


viite-UI/src/view/LinkPropertyLayer.js
```diff
@@ -761,6 +763,10 @@
           feature.roadLinkData = road;
           simulatedOL3Features.push(feature);
           afterTransferLinks.push(road);
+          var marker = cachedMarker.createMarker(feature.roadLinkData);
+          if(map.getView().getZoom() > zoomlevels.minZoomForDirectionalMarkers) {
+            simulatedRoadsLayer.getSource().addFeature(marker);
```
Campanha: instead of simulatedRoadsLayer shouldnt we use directionMarkerLayer?

Campanha: Edit: Reviewed with with dev, and should be simulatedRoadsLayer like it is now since the Siirra only simulates the after transfer. (Y)

PR #808, opened 2017-07-11
VIITE-565 Get length from VVH

plourenco90:


oappicgi:
Fairly sure tests have to be changed as well, but i will run them after fixture reset to confirm

plourenco90:
I did the fixture reset and ran all tests and it passed.

PR #809, opened 2017-07-11
Viite 563 roadtype of project table should be the same as the one in the form

Campanha:


Campanha:
@pedrosag  yep they want it like this for now.

PR #810, opened 2017-07-12
Viite 451 reserving new address to unaddressed link

oappicgi:
added new PUT to viiteapi for adding address to nonaddressed link to given project. Currently only checks that address is not reserved in roadaddress table. checks for project-link table is other task


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -48,7 +49,7 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
   val DrawRoadPartsOnly = 2
   val DrawPublicRoads = 3
   val DrawAllRoads = 4
-
+  def withDynTransaction[T](f: => T): T = OracleDatabase.withDynTransaction(f)
```
Campanha: Could have been avoided if you opened the transactions at service level

oappicgi: in the pass i have had problems with rollback in tests if i had withDynTransaction block inside the test, so decided to add API level. (main problem being that rollback did not rollback)

PR #814, opened 2017-07-13
Viite 571 define error message when reserved

oappicgi:
error message now shows roadnumber, roadpartnumber and project date (as in task description)
Also fixes bug where fetching roadaddressproject startdate was switched with createdate


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -126,8 +126,10 @@ object ProjectDAO {
   }
 
   def updateRoadAddressProject(roadAddressProject: RoadAddressProject): Unit = {
+
+   val test= roadAddressProject.startDate
```
Campanha: unused val?

PR #817, opened 2017-07-14
Viite 585 connect save button to backend

hestrompa:
Contains refactoring to VIITE-449 and VIITE-451

### DIFF
--- a/viite-UI/src/utils/backend-utils.js
+++ b/viite-UI/src/utils/backend-utils.js
@@ -143,6 +143,26 @@
         });
     });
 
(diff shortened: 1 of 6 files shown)


hestrompa:
All tests ok after merge master to the branch

viite-UI/src/utils/backend-utils.js
```diff
@@ -143,6 +143,26 @@
         });
     });
 
+    this.insertNewRoadLink = _.throttle(function(data, success, failure) {
+      var Json = {
+        linkIds : data[0],
```
oappicgi: I might be wrong but isn't it possible to use data.linkid instead of position based matching?

PR #818, opened 2017-07-17
Viite 426 check if new roadpart is already reserved

Campanha:
Check in project_link table for same roadNumber & roadPart but ignoring current project id
Show popup in UI after error in saving new road address in project


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -173,6 +173,19 @@ object ProjectDAO {
     Q.queryNA[String](query).firstOption
   }
 
+
+  def isNewRoadPartLinkUsed(roadNumber: Long, roadPartNumber: Long, projectId: Long) = {
+    val query =
+      s"""
+		select prol.id
+         from project_link prol
+         join lrm_position pos on prol.lrm_position_id = pos.id
```
oappicgi: where do we need lrm_position?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -60,8 +60,8 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     * @return
     */
   def checkNewRoadAddressNumberAndPart(roadNumber: Long, roadPart: Long, project :RoadAddressProject): Option[String] = {
-      val roadAddresses = RoadAddressDAO.isNewRoadPartUsed(roadNumber, roadPart, project.id)
-      if (roadAddresses.isEmpty) {
+      val projectLinks = ProjectDAO.isNewRoadPartLinkUsed(roadNumber, roadPart, project.id)
```
oappicgi: we also need to keep old check

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
pedrosag: Possibly to make sure that the Project_Link has a LRM_Possition connection therefore being a step closer to prove that it is correctly created (therefore being able to be used)?

Campanha: New version commit will not have this query. As Ossi said is more deeper than that, we need to keep the validation of:
1.new road address should not exist already in Road_address table
2.new road_address (same roadNumber and roadPart) should not exist in Project_link for others project ids

PR #821, opened 2017-07-18
DROTH-791 - Bus stops located in road registry while using VIITE

marquesfm:
- Modification at resolveAddressAndLocation method in GeometryTransform Class to fix the side_code value when geometry is digitazed to the opposite direction than the road address is growing.


leirasf:
It's ok to me

PR #822, opened 2017-07-18
Viite 577 flip project link sidecode from api

oappicgi:


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -231,6 +231,25 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
     }
   }
 
+
+  put("roadlinks/roadaddress/project/directionchangenewroadlink"){
```
hestrompa: The link should be "/roadlinks/roadaddress/project/directionchangenewroadlink". It's missing a "/" at the beggining of it.

plourenco90:
Changes applied on VIITE-576 pull request

PR #823, opened 2017-07-18
DROTH-793_complementary_links_not_drawn_after_zoom

oskarmalmstrom:
Created separate event listeners(set when the layer is created) to layers the check if complementaries are shown/not shown. The existing event listeners were not active when zoomed >200 meters.


leirasf:
Looks ok to me

joanacoelho:
ok

marquesfm:
Ok for me

PR #828, opened 2017-07-20
Viite 578 re calculating the m values

plourenco90:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -173,6 +173,11 @@ object ProjectDAO {
     }
   }
 
+  def updateMValues(projectLink: ProjectLink): Unit = {
+      sqlu"""update project_link set start_addr_m = ${projectLink.startAddrMValue}, end_addr_m = ${projectLink.endAddrMValue} where id = ${projectLink.id}
```
JaniKoivisto: Should we also update MODIFIED_DATE?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -173,6 +173,11 @@ object ProjectDAO {
...
+          """.execute
+        sqlu"""update lrm_position set start_measure = ${projectLink.startMValue}, end_measure = ${projectLink.endMValue}  where id = ${projectLink.lrmPositionId}""".execute
```
JaniKoivisto: I feel its unnecessary to update  START_MEASURE and END_MEASURE of linear location.

PR #830, opened 2017-07-21
Viite 600 chosen links not getting new road address value

Campanha:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectDeltaCalculator.scala
```diff
@@ -68,13 +71,59 @@ object ProjectDeltaCalculator {
     ).toSeq
   }
 
+  def orderProjectLinksTopologyByGeometry(list:Seq[ProjectLink]): Seq[ProjectLink] = {
+
+    def linkTopologyByGeomEndPoints(sortedList: Seq[ProjectLink], unprocessed: Seq[ProjectLink]) : Seq[ProjectLink] = {
```
oappicgi: I think there should be error state here. If i am not mistaken (and i might be) we force list to be sorted at the end.  It is possible that user has selection that we cannot sort, especially after we add feature where user can add links by pressing shift.

oappicgi:
Since it fixes issue in ticket i think we can merge it. I added task 606 for checking  non-connecting branching and throw user error if that is the case.

PR #831, opened 2017-07-25
VIITE-594 Showing project change table button after user has saved ne…

hestrompa:
…w road and there are no roads that need modification

Changed the logic to show the changeTable and to show the button to send values to TR.


oappicgi:
wrote in skype, but basically doesn't have room for publish button shown in picture.

PR #835, opened 2017-07-28
Viite 614 saving multiple times to the same road number and road part

hestrompa:
This pull request incluedes tasks: 
VIITE-614 Saving multiple times to the same road number and road part, updating the existant road address
VIITE-612 Extra calibration points appear in UI, when using the flip function


oappicgi:
👍 think it is ready to be merged, although having one test to test adding roadlink to project's new roadnumber & roadpart wouldn't hurt.  Think currently changed ones focus on flipping

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -217,17 +222,15 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     }).asInstanceOf[Seq[ProjectLink]]
   }
 
-  def changeDirection(projectLink:Seq[Long]): String = {
+  def changeDirection(projectId : Long, roadNumber : Long, roadPartNumber : Long): String = {
```
tvenhola: It'd be better to return Option[String] as an error message: None for no errors, otherwise Some("errormessage")

PR #841, opened 2017-08-03
Viite 568 creation of calibration points for the roadpart with trackcode

plourenco90:
VIITE-568


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -148,13 +148,20 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     }
   }
 
+  def getSideCode(links: Seq[ProjectAddressLink]): SideCode = {
```
tvenhola: Should be renamed: it's confusing the way it is now

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -152,6 +152,27 @@ object ProjectDAO {
     }
   }
 
+  def getProjectLinkById(projectLinkId: Long): List[ProjectLink] = {
```
tvenhola: should be "ByLinkId" as we're matching LRM_POSITION.link_id = $projectLinkId

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -464,6 +475,27 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
   }
 
+  def getProjectRoadLinksByLinkIdsByProject(linkIdsToGet : Set[Long], projectId: Long, newTransaction : Boolean = true): Seq[ProjectAddressLink] = {
```
tvenhola: Rename, needs better (shorter) name

tvenhola:
Refactoring and fixing ready. Comments?

PR #842, opened 2017-08-08
Viite 622 ely value  1 in change table

pedrosag:
*Fixed small styler bug, now suravage roadlink style has maximum priority.  
*When adding new road addresses (Ussi) to a project, it will now update the project ELY if said ELY is -1 to the ELY of the road. *It is not possible to add roads with different ELY's of the project, it will output a error message (the message itself needs to either be translated to Finnish or replaced to a suitable one.
*Adding test cases for the update of the ELY in service level and in DAO level.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -767,5 +768,24 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
+    val currentProjectEly = getProjectEly(currentProjectId)
+    if (currentProjectEly == -1) {
+      ProjectDAO.updateProjectEly(currentProjectId, newEly)
+      ""
+    } else if (currentProjectEly == newEly){
+      logger.info("ProjectId: " + currentProjectId + " Ely is \"" + currentProjectEly + "\" no need to update")
+      ""
+    } else {
+      logger.error("Cannot update project to another ELY, current road can't be added")
+      //TODO: translate or find a suitable message in Finnish
```
pedrosag: I just need a small clarification, Is this phrase:
"Virhe: Projektia ei voi päivittää toiselle ELY: lle, nykyistä tieä ei voi lisätä"
Ok or should it be switched to another?

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -314,12 +314,18 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
       val projectLink = parsedBody.extract[NewRoadAddressExtractor]
       val roadLinks = projectService.getProjectRoadLinksByLinkIds(projectLink.linkIds)
       withDynTransaction {
-        val errorMessage = projectService.addNewLinksToProject(roadLinks, projectLink.projectId, projectLink.newRoadNumber, projectLink.newRoadPartNumber, projectLink.newTrackCode, projectLink.newDiscontinuity)
-        if (errorMessage == "") {
-          Map("success" -> true)
+        val elyUpdateErrorMessage = projectService.setProjectEly(projectLink.projectId, projectLink.roadEly)
```
tvenhola: Let's not use strings and comparisons but Option[String] and None to mark success

PR #843, opened 2017-08-09
Viite 92 delete a road part from project

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -306,12 +306,19 @@ object ProjectDAO {
   }
 
 
-  def removeProjectLinksById(projectLinkIds: Set[Long])= {
+  def removeProjectLinksById(projectLinkIds: Set[Long], lrmPositionIds: Set[Long]) = {
```
tvenhola: We could delete the LRM position ids using a single set parameter: check ids for project links by querying, delete lrm positions by those ids, delete project links by those ids.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -322,6 +329,14 @@ object ProjectDAO {
     Q.updateNA(query).first
   }
 
+  def removeProjectLinksByProject(projectId : Long) = {
+    val query =
+      s"""
+         DELETE FROM Project_Link WHERE project_id = ${projectId}
```
tvenhola: Should also delete LRM positions

viite-UI/src/model/RoadAddressProjectCollection.js
```diff
@@ -336,6 +361,11 @@
       return dirtyProjectLinks;
     };
 
+    this.setCurrentRoadPartList = function(parts){
+      currentRoadPartList = parts.slice(0);
+      dirtyRoadPartList = parts.slice(0);
```
tvenhola: Why are they defaulting to dirty?

viite-UI/src/model/RoadAddressProjectCollection.js
```diff
@@ -142,7 +142,15 @@
         name: data[0].value,
         startDate: data[1].value,
         additionalInfo: data[2].value,
-        roadPartList: dirtyRoadPartList
+        roadPartList: _.map(dirtyRoadPartList, function(part){
```
tvenhola: We should use only one list but have `dirty: (true | false)` for them and handle them in post that way

hestrompa: Maybe to have only one list we need to have 2 booleans, because we need to see if it's a dirty deleted or a dirty new added.

viite-UI/src/model/RoadAddressProjectCollection.js
hestrompa: The dirty is saving all temporary changes in UI and is the list that the user can see. The currentRoadPartList is used when user presses Peruuta and wants to discard all changes in dirty list

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
tvenhola: I'll do the DAO refactorings

viite-UI/src/model/RoadAddressProjectCollection.js
hestrompa: And I think that have two lists is propably useful as well on task VIITE-616, where we need to show 2 different lists to user, and we can filter the dirty list there to show the unsaved road parts

tvenhola:
SBT tests not passed :|

tvenhola:
Okay, we keep the UI implementation as-is for now. Refactor it if we later have some advantage from doing it.

tvenhola:
Refactoring the ProjectService

PR #845, opened 2017-08-10
Viite 616 reserving a road part to an unfinished project

plourenco90:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -135,7 +135,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
         val ely: Option[Long] = roadLink.headOption.map(rl => MunicipalityDAO.getMunicipalityRoadMaintainers.getOrElse(rl.municipalityCode, 0))
         ely match {
           case Some(value) if value != 0 =>
-            Some(ReservedRoadPart(roadpartid, roadnumber, roadpart, lenght, Discontinuity.apply(discontinuity.toInt), value, Some(startDate), Some(endDate)))
+            Some(ReservedRoadPart(roadpartid, roadnumber, roadpart, lenght, Discontinuity.apply(discontinuity.toInt), value, Option(startDate), Option(endDate)))
```
tvenhola: Some(A) and Option(A) are interchangable

tvenhola:
Repeatedly clicking Varaa will create copies of the road part line in Form
image

tvenhola:
Saving a project with new created road parts will throw an error
image

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
plourenco90: The change here was because the Option's apply function makes a null into None. When the end address is null from the DB, and if you do Some(null) = Some(null), and if you do Option(null)= None.
But i will check if i can change it in the DAO instead of in the Service class.

oappicgi:
project change table is not updating, and reserver roads seem to go into wrong list

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -1041,6 +1041,6 @@ object RoadAddressDAO {
              on r.START_ADDR_M=ra.lol
              WHERE r.road_number=$roadNumber AND r.road_part_number=$roadPart AND
              (r.valid_from is null or r.valid_from <= sysdate) AND (r.valid_to is null or r.valid_to > sysdate) AND track_code in (0,1)"""
-    Q.queryNA[(Long,Long,Double,Long, DateTime, DateTime)](query).firstOption
+    Q.queryNA[(Long,Long,Double,Long, Option[DateTime], Option[DateTime])](query).firstOption
```
tvenhola: We may produce null pointer exceptions by this change

tvenhola: I am officially now an idiot.

PR #848, opened 2017-08-11
Viite 632 reversing address direction miscalculates m values

tvenhola:
VIITE-632/633, reversing address direction bug fixes


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -172,7 +172,16 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
+    val randomSideCode =
+      linksInProject.map(l => l -> projectAddressLinks.find(n => GeometryUtils.areAdjacent(l.geometry, n.geometry))).toMap.find { case (l, n) => n.nonEmpty }.map {
+        case (l, Some(n)) =>
+          if (GeometryUtils.areAdjacent(l.geometry.head, n.geometry.last) ||
+            GeometryUtils.areAdjacent(l.geometry.last, n.geometry.head))
+            l.sideCode
+          else
+            switchSideCode(l.sideCode)
+        case _ => SideCode.TowardsDigitizing
+      }.getOrElse(SideCode.TowardsDigitizing)
```
Campanha: could this "OrElse(SideCode.TowardsDigitizing)" be redundant since we already have "case _ => SideCode.TowardsDigitizing"?

tvenhola: if the map is empty then the result is None. Hence the getOrElse is needed.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -205,19 +214,28 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
   }
 
   def changeDirection(projectId : Long, roadNumber : Long, roadPartNumber : Long): Option[String] = {
+    RoadAddressLinkBuilder.municipalityRoadMaintainerMapping // make sure it is populated outside of this TX
```
tvenhola: We might want to populate this at startup.

PR #851, opened 2017-08-11
VIITE-637 Enabling button on link creation + small api refactoring

plourenco90:
After creating a new link and opening the change table is now enabling the publish button
Removed the transaction from the API to the Service class


tvenhola:
Looks good

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -175,48 +175,51 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
     }
     withDynTransaction {
```
oappicgi: think this is causing run with rollback to leave data behind

Campanha: After fixture reset its normal to have the project "ProjectOne" with road number 77997 and road part 1 which is in insert_project_link_data.sql.  Project_link in DB should be the same before and after   SBT tests

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -246,7 +246,7 @@ object ProjectDAO {
     }
     val query =
       s"""SELECT id, state, name, created_by, created_date, start_date, modified_by, modified_date, add_info, status_info, ely
-          FROM project $filter order by name, id """
+          FROM project $filter where created_date is not null and modified_date is not null and start_date is not null order by name, id """
```
tvenhola: This would filter out all unmodified projects from project listing. It'd be better to
a) return datetimes as option[datetime]s
b) change table definition and require created_date to be not null -> can't be null and return modified_date as COALESCE(modified_date, created_date)

tvenhola: created_date is non-null already, going for b)

Campanha: yes, created_date and start_date have that db constraint already.

PR #861, opened 2017-08-21
DROTH-833_Servicepoints_are_not_shown_on_map

MAC-CGI:
DROTH-833_Servicepoints_are_not_shown_on_map


leirasf:
It's OK to me. 
But maybe in the future we should start extending the point asset to have one for the point asset with complimentary links.

marquesfm:
It´s ok

PR #862, opened 2017-08-21
DROTH-830_Maximum_value_traffic_sign

oskarmalmstrom:
-TrafficSignLabel changed to show question mark when value of sign is over 120 (for sign types 1-5)
-Asset grouping distance radius is not retrieved from application model, but is passed from application.js


marquesfm:
ok

leirasf:
It's Ok

PR #863, opened 2017-08-22
Viite 428 holding shift button should allow add delete non addressed links

Campanha:


viite-UI/src/view/MapView.js
```diff
@@ -130,7 +131,18 @@
 
     //when the map dragging stops the cursor value returns to the initial one
     map.on('pointerup', function(evt) {
-      map.getViewport().style.cursor = "initial";
+      map.getViewport().style.cursor = oldCursorStyle;
+    });
+
+    eventbus.on('keydown', function(tool){
```
oappicgi: think this causes cursor constantly switch between  two cursor styles when adding links to selection.

Campanha: Now you say that, it is a nice question. 428 says cursor and selection should work by holding shift and clicking, while 440 says cursor should change when shift key is pressed

tvenhola: Let's fix this

PR #866, opened 2017-08-25
Viite 500 add as is

JaniKoivisto:


tvenhola:
Tests pass after master merged and conflicts resolved

PR #867, opened 2017-08-28
DROTH-865 Bus stops having last modified date vvh_generated even when…

marquesfm:
… link is not been updated EVER

and
DROTH-868 Adjust the last modified date for all point-like assets

-Modify the calculation of the Points of The asset, on base of LinearReference(mValue and geometry).


oskarmalmstrom:
OK

leirasf:
OK

PR #869, opened 2017-08-29
Viite 591 selecting a template which has already been modified

hestrompa:


viite-UI/src/view/ProjectLinkLayer.js
```diff
@@ -128,7 +128,8 @@
       layer: [vectorLayer, suravageRoadProjectLayer],
       condition: ol.events.condition.singleClick,
       style: function(feature, resolution) {
-        if (feature.projectLinkData.status === notHandledStatus || feature.projectLinkData.status === newRoadAddressStatus || feature.projectLinkData.roadLinkSource === 3){
+        if (feature.projectLinkData.status === notHandledStatus || feature.projectLinkData.status === newRoadAddressStatus ||
```
tvenhola: We should probably replace this with a utility method for checking the status. These status checks seem to get common and could be replaced with a function projectLinkStatusIn(projectLink, [allowedStatus1, allowedStatus2, ...]).

viite-UI/src/view/RoadAddressProjectEditForm.js
```diff
@@ -204,9 +205,17 @@
     };
 
     var changeDropDownValue = function (statusCode) {
+      var rootElement = $('#feature-attributes');
       if (statusCode === STATUS_UNCHANGED) {
         $("#dropDown").val('ennallaan').change();
       }
+      else if(statusCode === STATUS_NEW_LINK){
+        $("#dropDown").val('uusi').change();
```
tvenhola: We should work away the Finnish constant strings from the code and replace them with "static" variables or other kind of references.

PR #884, opened 2017-09-06
Viite 678 ennallaan roadlinks are changed into uusi roads

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -382,6 +382,15 @@ object ProjectDAO {
     Q.queryNA[Long](query).list
   }
 
+  def projectLinksCountUnchanged (projectId : Long, roadNumber:Long, roadPartNumber: Long): Long = {
+    var query =
+      s"""
+         select count(id) from project_link
+          WHERE project_id = $projectId and road_number = $roadNumber and road_part_number = $roadPartNumber and status = 4
```
tvenhola: Avoid magic numbers -> use constants from LinkStatus

hestrompa:
Adding unit tests

PR #885, opened 2017-09-06
Droth 836 add platform number field to bs properties

joanacoelho:


digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -360,6 +361,15 @@ Returns empty result as Json message, not as page not found
       halt(NotAcceptable("Invalid Mass Transit Stop direction"))
   }
 
+
+  private def validatePropertiesMaxSize(properties: Seq[SimpleProperty]) = {
+    val propertiesWithMaxSize: Map[String, Int] = massTransitStopService.getPropertiesWithMaxSize()
+    val invalidPropertiesDueMaxSize: Seq[SimpleProperty] = properties.filter { property =>
+      propertiesWithMaxSize.contains(property.publicId) && property.values.nonEmpty && property.values.forall { value => value.propertyValue.length > 3 }
```
joanacoelho: I changed this before this comment - it was not checking if the property was empty

joanacoelho:
Ok

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
leirasf: This is not validating for all the properties instead of for the propertiesWithMaxSize?

UI/src/view/MassTransitStopForm.js
```diff
@@ -49,6 +49,13 @@
     };
   };
 
+  var ValidatePlatformNumberMaxSize = function (target, propertyPublicId) {
```
leirasf: Convention the names should start with lowercase.

UI/src/view/MassTransitStopForm.js
```diff
@@ -263,6 +270,7 @@
           elementType = property.propertyType === 'long_text' ?
             $('<textarea />').addClass('form-control') : $('<input type="text"/>').addClass('form-control');
           element = elementType.bind('input', function(target){
+            ValidatePlatformNumberMaxSize(target, property.publicId);
```
leirasf: We are binding this for all the inputs!? Since we have the if for the properties probably it could be bind only when we are creating the platform input.

marquesfm:
Ok for me

PR #891, opened 2017-09-08
VIITE-692 - Project form becomes black and "locked" if user tries to select a link from the map

pedrosag:


viite-UI/src/view/LinkPropertyLayer.js
```diff
@@ -1317,15 +1317,15 @@
       deactivateSelectInteractions();
     });
 
-    eventListener.listenTo(eventbus, 'linkProperties:deactivateAllSelections', function(){
+    eventListener.listenTo(eventbus, 'linkProperties:deactivateAllSelections roadAddressProject:stopAllInteractions', function(){
```
tvenhola: maybe we should name them consistently "deactivateAllSelections".

PR #896, opened 2017-09-14
Viite 501 numbering change on transfer operation same road number

plourenco90:
Viite 501 numbering change on transfer operation same road number


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -375,6 +375,10 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     }
   }
 
+  private def isRoadPartTransfer(projectLinks: Seq[ProjectLink], roadNumber: Long , newRoadPart: Long): Boolean = {
+    projectLinks.filter(_.roadPartNumber == newRoadPart).length == 0 && projectLinks.filter(_.roadNumber == roadNumber).length > 0
```
tvenhola: .exists() is superior to .filter...length

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -673,29 +677,43 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
-        } else {
-          ProjectDAO.updateProjectLinkStatus(updatedProjectLinks.map(_.id).toSet, newStatus, userName)
         }
-        recalculateProjectLinks(projectId, userName)
+        case LinkStatus.Transfer => {
+          if (isRoadPartTransfer(updatedProjectLinks, newRoadNumber, newRoadPart)) {
+            val updated = updatedProjectLinks.map(updl => {
+              updl.copy(roadPartNumber = newRoadPart, status = linkStatus, calibrationPoints = (None, None))
+            })
+            ProjectDAO.updateProjectLinksToDB(updated.map(upd => upd), userName)
```
tvenhola: map(upd => upd) can be removed

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectDeltaCalculator.scala
```diff
@@ -114,26 +125,30 @@ object ProjectDeltaCalculator {
...
+        n.sortBy(_.startAddrMValue).groupBy(ra => (ra.roadNumber, ra.roadPartNumber, ra.track))
+          .mapValues(v => combine(v.sortBy(_.startAddrMValue))).values.flatten.map(ra =>
+          RoadAddressSection(ra.roadNumber, ra.roadPartNumber, ra.roadPartNumber,
+            ra.track, ra.startAddrMValue, ra.endAddrMValue, ra.discontinuity, ra.roadType)).toSeq
+      })
+
+    sourceCombined.map { sec =>
+      val linkId = roadAddresses.find(ra => sec.head.includes(ra)).map(_.linkId).get
+      val targetGroup = targetCombined.find(tc => tc.head.includes(projectLinks.find(_.linkId == linkId).get))
+      sec.head -> targetGroup.get.head
```
tvenhola: why sec.head? What happens to other elements?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
```diff
@@ -67,7 +67,7 @@ object ProjectSectionCalculator {
     */
   def assignMValues(projectLinks: Seq[ProjectLink]): Seq[ProjectLink] = {
     val (terminated, others) = projectLinks.partition(_.status == LinkStatus.Terminated)
-    val (newLinks, nonTerminatedLinks) = others.partition(_.status == LinkStatus.New)
+    val (newLinks, nonTerminatedLinks) = others.partition(l => l.status == LinkStatus.New || l.status == LinkStatus.Transfer)
```
tvenhola: Transfer shouldn't be given completely new addresses. (may produce incorrect results if address length doesn't match to geometry length)
nonTerminatedLinks should be renamed as addressedLinks for clarity.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
tvenhola: What is the check that this method actually does?
There is a link in this sequence where this newRoadPart is not mentioned but this roadNumber is?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectDeltaCalculator.scala
plourenco90: Because that list always contain one element with a seq, anyway the list is now flatten so there is not need to .head

oappicgi:
tested it visually: first made transfer from 206 to 207 which seemed ok. then each time i applied unchanged to 206 i got warning that roadpart has already been reserved("tämä tieosoite on jo käytössä"), but it applied changed anyway.  When i tried to apply transfer to rest of the 207 road  it gives same error, but this time refuses to apply changes. Someone should retest this before merge.

plourenco90:
Made the exact same steps, this was the result in mine
e6b7eb8a-50d6-48ac-8746-32f4272cfebc

oappicgi:
in your pic, last row seems odd. To my understanding last row should be  Siirto/transfer (because  m values have changed due to adding link to that roadpart)  AET on target side on same row seems also wrong, i believe it should be 69 since  siirto row is 0-69

PR #898, opened 2017-09-14
Viite 680 + 683 (from 466) copying project to road address table after receiving TR save message & tries to form geometry if we are lacking some

oappicgi:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -965,12 +965,12 @@ object RoadAddressDAO {
         case None => ""
       })
       addressPS.setString(11, createdBy.getOrElse(address.modifiedBy.getOrElse("-")))
-      val (p1, p2) = (address.geometry.head, address.geometry.last)
-      addressPS.setDouble(12, p1.x)
-      addressPS.setDouble(13, p1.y)
+      val (p1, p2) = (address.geometry.headOption, address.geometry.lastOption)
```
tvenhola: Option seems useless as we call .get after this every time.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -908,20 +910,104 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
+      if (projectLinks.isEmpty)
+        throw new RuntimeException(s"Tried to import empty project to road address table after TR response : $newState")
+
+      val roadAddressIDsToExpire= RoadAddressDAO.fetchByLinkId(projectLinks.map( x => x.linkId ).toSet,floatingFalse,historyFalse)
+      //Expiring all old addresses by their ID
+      roadAddressService.expireRoadAddresses(roadAddressIDsToExpire.map( x => x.id ).toSet)
+      //Create endDate rows for old data that is "valid" (row should be ignored after end_date)
+      RoadAddressDAO.create(roadAddressIDsToExpire.map(x=> x.copy(endDate =projectStartDate,id =NewRoadAddress)),Some(project.head.createdBy))
+      //Create new rows to RoadAddress table defining when new address is used
+      importProjectLinksToRoadAddressTable(projectLinks.map(x=>x.copy(endDate = None, startDate =projectStartDate)),roadAddressIDsToExpire,Some(project.head.createdBy))
```
tvenhola: For terminated road addresses we should not have anything without an end date.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
oappicgi: ah, yea forgot to remove those as well, had logic to add null but DB wouldn't accept it

PR #900, opened 2017-09-18
Viite 705+711 bug in combination of actions uusi siirto ennallaan and growth direction

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -184,9 +184,13 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
       }
       if (GeometryUtils.areAdjacent(newLink.geometry.head, endP) ||
         GeometryUtils.areAdjacent(newLink.geometry.last, startP))
-        SideCode.TowardsDigitizing
-      else
-        SideCode.AgainstDigitizing
+        if (existingLink.sideCode.equals(AgainstDigitizing))
```
tvenhola: This looks suspicious: the adjacency check looks for newLink start being at the end of the previous link or vice versa, but sidecodes don't seem to match.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
```diff
@@ -194,8 +194,8 @@ object ProjectSectionCalculator {
             })
       }
     }
-    val groupedProjectLinks = newProjectLinks.groupBy(record => (record.roadNumber, record.roadPartNumber))
-    val groupedOldLinks = oldProjectLinks.groupBy(record => (record.roadNumber, record.roadPartNumber))
+    val groupedProjectLinks = newProjectLinks.sortBy(_.startAddrMValue).groupBy(record => (record.roadNumber, record.roadPartNumber))
```
tvenhola: better to do mapValues(_.sortBy...) and not sort before grouping.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
```diff
@@ -232,6 +232,7 @@ object ProjectSectionCalculator {
         case (_, Some(LinkStatus.UnChanged), _) => nextSection.map(_.startAddrM)
         case (LinkStatus.NotHandled, _, Some(value)) => Some(value + currentSection.endAddrM - currentSection.startAddrM)
         case (LinkStatus.Transfer, _, Some(value)) => Some(value + currentSection.endAddrM - currentSection.startAddrM)
+        case (LinkStatus.Numbering, _, Some(value)) => Some(value + currentSection.endAddrM - currentSection.startAddrM)
```
tvenhola: Numbering should just have currentSection.endAddrM as it shouldn't affect the values at all.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
tvenhola: With the code above, it looks like we should be using the existingLink and reverse the direction if needed. A thorough unit test should help with this.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
hestrompa: I'll change this

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
hestrompa: I'll change this

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
hestrompa: I think yes, we should use the existing value and reverse when we need. I'm just not sure if this validation is the most correct one, but I did some tests and it worked. I'll add a unit test

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectDeltaCalculator.scala
```diff
@@ -107,8 +111,9 @@ object ProjectDeltaCalculator {
   }
 
   def pair(roadAddress: Seq[RoadAddress], projectLink: Seq[ProjectLink]): Seq[(RoadAddress,ProjectLink)] = {
+    val df = new DecimalFormat("#.###")
     roadAddress.foldLeft(List.empty[(RoadAddress,ProjectLink)]) { case (p, a) =>
-      projectLink.find(b => a.linkId == b.linkId) match {
+      projectLink.find(b => a.linkId == b.linkId && b.endMValue == df.format(a.endMValue - a.startMValue).replace(",", ".").toDouble ) match {
```
tvenhola: comparison between an endMValue of b and then segment length of a doesn't make sense.
Double comparison with == is no good. Precision should be used instead.
fi.liikennevirasto.viite.MaxAllowedMValueError

for example:
Math.abs(a.endMValue - b.startMValue) <= MaxAllowedMValueError

A new user story, 719, will address this pairin issue.

hestrompa: The comparison between the endMValue of b and the length of a is to compare the segments when the link id have more than one address on road_address table. We cannot assume that values on project_link, because the mValues are not the same in Project_Link and Road_address table.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectSectionCalculator.scala
hestrompa: I think this is not necessary anymore

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/ProjectDeltaCalculator.scala
Campanha: @tvenhola in 
Math.abs(a.endMValue - b.startMValue) <= MaxAllowedMValueError should be
(Math.abs(a.endMValue - b.startMValue) - MaxAllowedMValueError )<= MaxDistanceDiffAllowed
;with  MaxAllowedMValueError  =>  0.001; MaxDistanceDiffAllowed => 1.0 ?

PR #904, opened 2017-09-25
DROTH-913 Bus stops: Ylläpitäjä as mandatory field: 

joanacoelho:
The field Ei tiedossa will be removed from the form in DROTH-912


leirasf:
Ok

marquesfm:
Ok for me

PR #912, opened 2017-09-28
Viite 709 master styler for the project data screen

plourenco90:


tvenhola:
Looks really good.

oappicgi:
otherwise good, but unknown links seem to have wider strokes that they are suppose to
stroke

PR #914, opened 2017-10-02
Viite 710 display only the newest road address data on link property layer

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -526,7 +529,7 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
 
     val addresses =
       withDynTransaction {
-        RoadAddressDAO.fetchByLinkId(roadLinksWithComplimentary.map(_.linkId).toSet, false, false).groupBy(_.linkId)
+        RoadAddressDAO.fetchByLinkId(roadLinksWithComplimentary.map(_.linkId).toSet, false, false).filterNot(ra => ra.endDate.isDefined).groupBy(_.linkId)
```
tvenhola: It's better to do the filtering in a query and not to fetch any database rows for filtering in the application code.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -139,9 +139,10 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
       withTiming(
         Await.result(combinedFuture, Duration.Inf), "End fetch vvh road links in %.3f sec"
       )
+    val roadsWithEndDate = addresses.values.flatten.filter(a => a.endDate.isDefined).map(_.linkId).toSeq
```
tvenhola: Could be fetched from the database separately with a query or could use .partition to split the addresses data into two separate values.
val (roadsWithoutEndDate, roadsWithEndDate) = addresses.values.flatten.partition(_.endDate.isDefined).map(_.linkId)

hestrompa: I cannot make this change on the query, because I need to filter the ones from VVH Roadlinks also. If I make the change on the query, instead of disappear, we just have 'missingRoadAddresses' on the map. I'll use the partitioner instead

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
hestrompa: I'll remove this validation, since we already have a boolean in the method fetchByLinkId that indicates if we want history of the road address and validates if the end date is null

PR #915, opened 2017-10-02
DROTH-933 Related to DROTH-809, Roadlink is shown red even without E-…

joanacoelho:
…road number :

-Validation for not insert assets when TR value is empty or null


joanacoelho:
ok

PR #916, opened 2017-10-03
DROTH-934 Cutting tool is not working properly with speed limit asset:

joanacoelho:
- Add function for speedLimit:unselect event
- Correction for cut without changes and correction for cut with changes


leirasf:
OK

PR #919, opened 2017-10-04
VIITE-743 - UI: changes to form

pedrosag:
*Added a Red "!" sign on the form (invisible before editing)
*Said sign goes visible if the user changes the endDistance (LOPUSSA) value to a value different of the original
*Restoring the Original value will remove said sign
*Added a parameter on the data to send to backend called userDefinedEndAddressM (is initially sent as -1)
*If the endAddressM is changed at the end then the userDefinedEndAddressM is changed to that
*Backend is reading said value

### DIFF
--- a/viite-UI/src/model/RoadAddressProjectCollection.js
+++ b/viite-UI/src/model/RoadAddressProjectCollection.js
@@ -241,7 +241,7 @@
         roadEly: Number($('#roadAddressProject').find('#ely')[0].value),
         roadLinkSource: Number(_.first(changedLinks).roadLinkSource),
         roadType: Number($('#roadAddressProject').find('#roadTypeDropDown')[0].value),
-        userDefinedEndAddressM: -1
+        userDefinedEndAddressM: null
       };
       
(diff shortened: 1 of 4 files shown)


viite-UI/src/model/RoadAddressProjectCollection.js
```diff
@@ -240,11 +240,22 @@
         discontinuity: Number($('#roadAddressProject').find('#discontinuityDropdown')[0].value),
         roadEly: Number($('#roadAddressProject').find('#ely')[0].value),
         roadLinkSource: Number(_.first(changedLinks).roadLinkSource),
-        roadType: Number($('#roadAddressProject').find('#roadTypeDropDown')[0].value)
+        roadType: Number($('#roadAddressProject').find('#roadTypeDropDown')[0].value),
+        userDefinedEndAddressM: -1
```
tvenhola: Why magic value -1?

pedrosag: Honestly I thought it would be a way for backend to identify if there are, or not, user defined calibration points.

PR #926, opened 2017-10-09
VIITE-741 Backend: save user given calibration points to database table

pedrosag:


digiroad2-oracle/src/main/resources/db/migration/V128__recreating_calibration_point_table.sql
```diff
@@ -0,0 +1,25 @@
+Drop Table CALIBRATION_POINT;
```
oappicgi: wouldn't this drop all current calibration points?

pedrosag: Not in this specific context since this table was not being used in any way, the current calibration points are calculated on demand based on the entry CALIBRATION_POINTS on the PROJECT_LINK table.

oappicgi: oh, yea should have thought that bit more =)

PR #931, opened 2017-10-10
Viite 744 custom calibration points to map

oappicgi:
svg file for customcalibrationpoints and logic to check if customcalibrationpoint icon should be applied


viite-UI/src/map/CalibrationPointMarker.js
```diff
@@ -22,8 +21,17 @@
...
 
-    marker.setStyle(calibrationPointMarkerStyle);
-    return marker;
+      var calibrationPointMarkerStylecustom = new ol.style.Style({
+        image: new ol.style.Icon({
+          src: "images/custom-calibration-point.svg",
+          anchor: [0.5, 1]
+        })
+      });
+      if(data.calibrationCode<=firstCustomCalibrationPointValue)
```
plourenco90: Isn't this always true?

oappicgi: 1 & 2 should be automatically generated calibrationpoints, but you are right, 5 is probably higher value than it needs to be. Couldn't find values written. i will lower it to 4

PR #936, opened 2017-10-12
VIITE-767 changes to project link partitioning logic

tvenhola:
grouping of links by being in project or not and subgroup rules
applied the 0.1 meter connectedness rule

### DIFF
--- a/digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/GraphPartitioner.scala
+++ b/digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/GraphPartitioner.scala
@@ -8,8 +8,8 @@ import org.geotools.graph.structure.basic.BasicEdge
 import scala.collection.JavaConversions._
 
 trait GraphPartitioner {
-  protected def clusterLinks[T <: PolyLine](links: Seq[T]): Seq[Graph] = {
-    val generator = new BasicLineGraphGenerator(0.5)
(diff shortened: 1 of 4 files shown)


oappicgi:
http://localhost:9001/api/viite/project/roadlinks?zoom=12&bbox=546243.9394773282,6759817.977519371,547208.9394773282,6760886.977519371&id=7091389 still results nullpointer exception for me in vvhclient 622

tvenhola:
This presumably happens because some of the suravage links are duplicated in VVH.

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/linearasset/GraphPartitioner.scala
```diff
@@ -8,8 +8,8 @@ import org.geotools.graph.structure.basic.BasicEdge
 import scala.collection.JavaConversions._
 
 trait GraphPartitioner {
-  protected def clusterLinks[T <: PolyLine](links: Seq[T]): Seq[Graph] = {
-    val generator = new BasicLineGraphGenerator(0.5)
+  protected def clusterLinks[T <: PolyLine](links: Seq[T], tolerance: Double = 0.5): Seq[Graph] = {
```
Campanha: should we use fi.liikennevirasto.viite.MaxSuravageToleranceToGeometry for tolerance value?

tvenhola: Partitioner can use any given value, but for viite the "connectedness" is defined as < 0.1 meters. (It's used in projectlinkpartitioner, for example)

PR #937, opened 2017-10-12
VIITE-748 - Poistu-painikkeessa häiriö / Poistu-button has a bug

pedrosag:
*Changed Cancel behaviour when editing an already opened project
*Added display of confirmation message
*Enforced an wait period on opening the project selection screen after a save is triggered

### DIFF
--- a/viite-UI/src/view/RoadAddressProjectForm.js
+++ b/viite-UI/src/view/RoadAddressProjectForm.js
@@ -3,7 +3,7 @@
     var currentProject = false;
     var selectedProjectLink = false;
     var activeLayer = false;
-    var hasNewRoadParts = false;
+    var hasReservedRoadParts = false;
     var staticField = function(labelText, dataField) {
       var field;
       field = '<div class="form-group">' +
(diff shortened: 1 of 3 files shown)


viite-UI/src/view/RoadAddressProjectForm.js
```diff
@@ -557,19 +557,21 @@
       });
 
       rootElement.on('click', '#cancelEdit', function(){
-        reOpenCurrent();
+        new GenericConfirmPopup('Haluatko tallentaa tekemäsi muutokset?', {
+          successCallback: function() {
```
hestrompa: I didn't tested yet, but I don't know if the logic is correct here (at least looking at the names of functions doesn't seem correct). When we show the popup with the "Haluatko tallentaa tekemäsi muutokset?", we will close the project anyway, independently of the user choice. The only difference between the buttons is that one "save the changes and exit" and the other one just "exit the project without save". None of the options should go "next and enable interactions"

hestrompa: I didn't tested yet, but I don't know if the logic is correct here (at least looking at the names of functions doesn't seem correct). When we show the popup with the "Haluatko tallentaa tekemäsi muutokset?", we will close the project anyway, independently of the user choice. The only difference between the buttons is that one "save the changes and exit" and the other one just "exit the project without save". None of the options should go "next and enable interactions"

pedrosag: 2 behaviours:

1. on success it will save changes and load the action form

2. on close it will just load the action form discarding changes.

oappicgi:
seems to work if "user plays nice". If user zooms map  (at least with mouse click) save will be terminated.

PR #939, opened 2017-10-13
VIITE-TRMock always reads values as non null

oappicgi:
instead of checking key we check key value

### DIFF
--- a/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteTierekisteriMockApi.scala
+++ b/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteTierekisteriMockApi.scala
@@ -79,13 +79,10 @@ class ViiteTierekisteriMockApi extends ScalatraServlet with JacksonJsonSupport {
     }
 
     def testIsNull(map: Map[String, Any], keys: Seq[String], errorTemplate: String): Unit = {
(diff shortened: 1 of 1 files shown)


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteTierekisteriMockApi.scala
```diff
@@ -79,13 +79,10 @@ class ViiteTierekisteriMockApi extends ScalatraServlet with JacksonJsonSupport {
...
-      if (keys.nonEmpty) {
-        val key = keys.head
-        val errorMessage = errorTemplate.format(key)
-        if (map(key) != "null") {
-          halt(failValidation(errorMessage))
-        }
-        testIsNull(map, keys.tail, errorTemplate)
+      val nonNullValues=map.filterNot(x=>x._2==null)
+      if (nonNullValues.nonEmpty) {
+        val errorMessage = errorTemplate.format(nonNullValues.head._1)
```
Campanha: is head of list enough? or should need a interaction for all nonNullValues?

oappicgi: also considered that, but since it is just our test api i thought showing first error is enough. Think TR will also stop on first error it finds.

PR #940, opened 2017-10-13
VIITE-770 Fix incorrect error message on link update

tvenhola:
failed mValue calculation should not be an error on update
show API returned error message in UI


tvenhola:
Fast merging

PR #949, opened 2017-10-19
VIITE-796 - "Etäisyyslukema"

pedrosag:
*Ended the oppression of the suravage links, they can now bear calibration points

### DIFF
--- a/viite-UI/src/view/ProjectLinkLayer.js
+++ b/viite-UI/src/view/ProjectLinkLayer.js
@@ -728,6 +728,8 @@
       var separated = _.partition(projectCollection.getAll(), function (projectRoad) {
         return projectRoad.roadLinkSource === 3;
       });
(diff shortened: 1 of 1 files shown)


viite-UI/src/view/ProjectLinkLayer.js
```diff
@@ -728,6 +728,8 @@
       var separated = _.partition(projectCollection.getAll(), function (projectRoad) {
         return projectRoad.roadLinkSource === 3;
```
Campanha: lets use __LinkGeomSource.SuravageLinkInterface.value__ instead __3__

pedrosag:
In order to see it in action:

- Open up the project
- Go to here: 6747792, 421583
- Create a project there
- Use "new" on both Suravage links, top one with 4; 118; 2, bottom one with 4; 118; 1
- Save them both

Calibration points now apear

PR #951, opened 2017-10-23
Viite 798 fix split addressing

tvenhola:
Fixes for incorrect split point being passed to the backend, negative values for resplitting (currently refuses resubmitting) and incorrect geometry for split links in address link builder.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
@@ -415,7 +415,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
   def splitSuravageLinkInTX(linkId:Long, username:String,
                             splitOptions: SplitOptions): Option[String] = {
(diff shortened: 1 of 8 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -415,7 +415,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
   def splitSuravageLinkInTX(linkId:Long, username:String,
                             splitOptions: SplitOptions): Option[String] = {
-    val sOption = getProjectSuravageRoadLinksByLinkIds(Set(linkId)).headOption
+    val sOption = getProjectSuravageRoadLinksByLinkIds(Set(Math.abs(linkId))).headOption
```
Campanha: is Math.abs really necessary for linkId?

tvenhola: We may get the negative link Id after the split if user re-splits the suravage link

oappicgi:
otherwise seems good, but i am still getting links to wrong direction

PR #952, opened 2017-10-23
VIITE-494 Rotating tr-id to enable resending project

oappicgi:
Contains tasks VIITE- 780 & 789

Instead of sending VIITE's project-id to TR we start sending TR_ID that we can easily change after it has been used in TR. We also query tierekisteri with tr_id instead of project_id. UI project listing shows orange "avaa uudelleen" which puts project to unfinnished state and clears tr_id field to null.


digiroad2-oracle/src/main/resources/db/migration/V133__new_TR_id_attribute.sql
```diff
@@ -0,0 +1 @@
+  ALTER TABLE PROJECT ADD tr_id NUMBER;
```
tvenhola: Should update all project tr_ids where status is not "in progress"

viite-UI/src/model/RoadAddressProjectCollection.js
```diff
@@ -542,5 +542,13 @@
         getData: getData
       };
     };
+
+    this.deleteRotatingProjectId = function(projectId){
```
tvenhola: Should have more sensible naming, like "reOpenProject".

viite-UI/src/model/RoadAddressProjectCollection.js
```diff
@@ -542,5 +542,13 @@
...
+      backend.reOpenProject(projectId, function(successObject) {
+        eventbus.trigger("roadAddressProject:rotatingIdDeleted",successObject);
```
tvenhola: More sensible naming for event: "roadAddressProject:reOpened"?

viite-UI/src/view/navigationpanel/ProjectListModel.js
```diff
@@ -56,26 +57,46 @@
...
           $('#project-list').html($(html));
           $('[id*="open-project"]').click(function(event) {
-            projectCollection.getProjectsWithLinksById(parseInt(event.currentTarget.value)).then(function(result){
-              setTimeout(function(){}, 0);
-              eventbus.trigger('roadAddress:openProject', result);
-              if(applicationModel.isReadOnly()) {
-                $('.edit-mode-btn:visible').click();
-              }
-            });
+            if(event.target.id.startsWith("reopen")){
```
tvenhola: Dangerous.

viite-UI/src/model/RoadAddressProjectCollection.js
```diff
@@ -542,5 +542,13 @@
...
+      }, function(errorObject){
+        console.log("Error at deleting rotatingId: " + errorObject);
```
tvenhola: Should show an error to the end user, with a modal.

PR #955, opened 2017-10-23
Viite 700 editing continuity and road type for project link

plourenco90:


plourenco90:
Can´t be merged before the VIITE-779

digiroad2-oracle/src/main/resources/db/migration/V131__add_required_attributes.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE PROJECT_LINK ADD ELY NUMBER;
```
tvenhola: Can't replace a migration file that is already in master

plourenco90: Ok, i will change it for V132

tvenhola:
Merging master into this branch

PR #961, opened 2017-10-24
VIITE-791 Prevent modification of project links

plourenco90:


pedrosag:
@oappicgi check on master class LinkValues.js, there should be there the project status already.

viite-UI/src/view/RoadAddressProjectEditForm.js
```diff
@@ -3,6 +3,10 @@
     var LinkStatus = LinkValues.LinkStatus;
     var LinkGeomSource = LinkValues.LinkGeomSource;
     var CalibrationCode = LinkValues.CalibrationCode;
+    var projectIncomplete = 1;
```
oappicgi: if we have there already defined somewhere like @pedrosag  said, then we should use them instead

PR #964, opened 2017-10-27
VIITE-818 - Discontinuity value on form is wrong when reselecting homogenized links

pedrosag:
*Added a selection of discontinuity to the form;
*Corrected a potential problem with the withGeometry method
*Created analogous methods of Guestimate but with ProjectLinks


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/GuestimateGeometryForMissingLinks.scala
```diff
@@ -100,4 +101,101 @@ class GuestimateGeometryForMissingLinks {
       }
     }
   }
+
```
tvenhola: We could eliminate duplicate code and have a generic type with some match cases for different subtypes

pedrosag: Good idea

tvenhola: This looks like a good implementation.
image

PR #971, opened 2017-10-31
Droth 960 more traffic signs to be implemented

marquesfm:


UI/src/view/MassTransitStopForm.js
```diff
@@ -121,6 +121,7 @@
       poistaSelected = false;
       var streetViewHandler;
       var isTRMassTransitStop = false;
+      var typeId = 10;
```
joanacoelho: this can be removed

UI/src/view/TrafficSignLabel.js
```diff
@@ -25,47 +25,116 @@
...
+        {signValue: [37], image: 'images/traffic-signs/leftBendSign.png'},
+        {signValue: [38], image: 'images/traffic-signs/severalBendRightSign.png'},
+        {signValue: [39], image: 'images/traffic-signs/severalBendLeftSign.png'},
+        {signValue: [40], image: 'images/traffic-signs/dangerousDescentSign.png'},
+        {signValue: [41], image: 'images/traffic-signs/steepAscentSign.png'},
+        {signValue: [42], image: 'images/traffic-signs/unevenRoadSign.png'},
+        {signValue: [43], image: 'images/traffic-signs/childrenSign.png'}
+      ];
+
+      function find() {
```
leirasf: Instead of having a method here you can have it as `var` and perform the find just once and then use it in all the methods below.

oskarmalmstrom:
ok

marquesfm:
Ok

PR #972, opened 2017-10-31
Viite 811 temporarily remove unknown from qa environment

hestrompa:


oappicgi:
BE: test("Reserve road part, renumber it and revert it; reservation should be freed")  fails

PR #981, opened 2017-11-03
DROTH-997: hide x-delete button when form is only showing once service.

oskarmalmstrom:
-checks current layer and amount of services for asset. Hide if only one service to prevent saving "empty" asset.

### DIFF
--- a/UI/src/view/PointAssetForm.js
+++ b/UI/src/view/PointAssetForm.js
@@ -9,10 +9,9 @@
     eventbus.on('application:readOnly', function(readOnly) {
       if(applicationModel.getSelectedLayer() == layerName && (!_.isEmpty(roadCollection.getAll()) && !_.isNull(selectedAsset.getId()))){
         toggleMode(rootElement, (editConstrains && editConstrains(selectedAsset)) || readOnly);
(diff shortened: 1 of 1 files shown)


leirasf:
Ok, just add the comment to add it in the "asset-type-layer-specifications.js" in the future.

UI/src/view/PointAssetForm.js
```diff
@@ -7,8 +7,14 @@
     var rootElement = $('#feature-attributes');
 
     eventbus.on('application:readOnly', function(readOnly) {
-      if(applicationModel.getSelectedLayer() == layerName && (!_.isEmpty(roadCollection.getAll()) && !_.isNull(selectedAsset.getId())))
+      if(applicationModel.getSelectedLayer() == layerName && (!_.isEmpty(roadCollection.getAll()) && !_.isNull(selectedAsset.getId()))){
         toggleMode(rootElement, (editConstrains && editConstrains(selectedAsset)) || readOnly);
+        if (layerName == 'servicePoints'){
+          if(isSingleService(selectedAsset)){
```
joanacoelho: This could be done in the if above

joanacoelho:
ok

PR #984, opened 2017-11-07
Viite 575 road addressing roundabouts

tvenhola:
Roundabout and ramp addressing:

* Roundabouts supporting single click addressing
* Ramps will use traffic direction from VVH if given and not connecting to any existing road part
* Ordinary calculation is used if no traffic direction is given and no existing road parts are connecting


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -233,45 +245,37 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
-              matchSideCodes(n, l)
-            case _ => SideCode.TowardsDigitizing
-          }.getOrElse(SideCode.TowardsDigitizing)
         val project = getProjectWithReservationChecks(projectId, newRoadNumber, newRoadPartNumber)
         if (!project.isReserved(newRoadNumber, newRoadPartNumber))
           ProjectDAO.reserveRoadPart(project.id, newRoadNumber, newRoadPartNumber, project.modifiedBy)
         val newProjectLinks = newLinks.map(projectLink => {
           projectLink.linkId ->
-            newProjectLink(projectLink, project, randomSideCode, newTrackCode, newRoadNumber, newRoadPartNumber,
+            newProjectLink(projectLink, project, SideCode.Unknown, newTrackCode, newRoadNumber, newRoadPartNumber,
```
oappicgi: i might be wrong, but didnt we ave randomsidecode here because we wanted to have default sidecode value  so it would be correct 50% of the time instead of ~0% of the time. (decreases times user has to switch sidecode)

tvenhola: the sideCode Unknown is corrected in recalculation. It guarantees the sidecodes to be set if we can process the road part. Uknown side code links don't participate in figuring out the direction and therefore they are preferred.

PR #986, opened 2017-11-08
Viite 483 flip addressing direction for parts in renumbering

Campanha:


Campanha:
all Backend end tests passing

tvenhola:
Delta calculator is still wrong after fixing the address values setting in DAO.

PR #991, opened 2017-11-09
Viite 854 numbering roads appear in reserved list

pedrosag:


tvenhola:
I think only the UI should change in that case and it should not show the true reserved list.

Or we should rework the whole reservation + ely code thingy

pedrosag:
@tvenhola The problem of doing this sort of thing in the UI is that in order to get the correct data in order to filter it there, I need to query both the PROJECT_RESERVED_ROAD_PART tables and the PROJECT_LINK table in order to find the ones whose LinkStatus is Numbering, which means another backend call already inside a response of said backend call.

Picking on said 1st idea, one could edit the response object of get: "/roadlinks/roadaddress/project/all/projectId/:id" in order to return both objects:

- one from the PROJECT_RESERVED_ROAD_PART

- the other from PROJECT_LINK

And then continuing the filter on the UI side as well.

Or, another feasible idea would be to change the PROJECT_RESERVED_ROAD_PART table to include the LinkStatus of said reservation. Filtering said results would be trivial in that case, but the rest of the scripts and case classes would have to be re-viewed.

oappicgi:
I would prefer to have boolean "numbering" attribute appended to class we get from DAO  which we could use to filter in UI. But this is fine as well.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -409,7 +409,7 @@ object ProjectDAO {
     }
   }
 
-  def getRoadAddressProjects(projectId: Long = 0): List[RoadAddressProject] = {
+  def getRoadAddressProjects(projectId: Long = 0, filterNumeric: Boolean = false): List[RoadAddressProject] = {
```
tvenhola: Let's call this renumbering, not numeric. Or filterNot: Set[LinkStatus]

Campanha:
i think in the future would be better to change reservation. But for now its good to use this workaround. After changes should be fine

tvenhola:
Let's go with the renaming and rework the reservations and releasing them in next sprint. Already spoke with PO about the challenges with transfers to other road parts or previously non-existent parts.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -418,9 +418,21 @@ object ProjectDAO {
       s"""SELECT id, state, name, created_by, created_date, start_date, modified_by, COALESCE(modified_date, created_date), add_info, status_info, ely
           FROM project $filter order by name, id """
     Q.queryNA[(Long, Long, String, String, DateTime, DateTime, String, DateTime, String, Option[String], Option[Long])](query).list.map {
-      case (id, state, name, createdBy, createdDate, start_date, modifiedBy, modifiedDate, addInfo, statusInfo, ely) =>
-        RoadAddressProject(id, ProjectState.apply(state), name, createdBy, createdDate, modifiedBy, start_date,
-          modifiedDate, addInfo, fetchReservedRoadParts(projectId), statusInfo, ely)
+      case (id, state, name, createdBy, createdDate, start_date, modifiedBy, modifiedDate, addInfo, statusInfo, ely) => {
+
+        if(!filterNotStatus.isEmpty) {
```
hestrompa: I think this logic is not clean enough. Maybe just changing it to if(filterNotStatus.nonEmpty) makes it simpler, but still the name of the variable filterNotStatus looks a little bit suspicious.

hestrompa: I think this logic is not clean enough. Maybe just changing it to if(filterNotStatus.nonEmpty) makes it simpler, but still the name of the variable filterNotStatus looks a little bit suspicious.

oappicgi: maybe just filtteredStatus or something so we don't have double negatives

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -291,7 +291,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     * @param linkIds the linkIds to process
     * @param roadNumber the roadNumber to apply/was applied to said linkIds
     * @param roadPartNumber the roadPartNumber to apply/was applied to said linkIds
-    * @param newLinks the whole newProjectLinks
+    * @param existingLinks the whole newProjectLinks
```
Campanha: existingLinks the whole *existing* ProjectLinks :p

pedrosag: Are you implying that newProjectLinks DO NOT EXIST GOOD SIR?!

tvenhola:
I'll fix that query filter thing

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -383,14 +383,27 @@ object ProjectDAO {
...
 
-  def fetchReservedRoadParts(projectId: Long): Seq[ReservedRoadPart] = {
-    Q.queryNA[(Long, Long, Long, Double, Long, Int, Long, Option[Long], Option[Long])](
-      "SELECT id, road_number, road_part_number, road_length, ADDRESS_LENGTH, discontinuity, ely, first_link_id," +
-        s"(select PROJECT_LINK.ID from PROJECT_LINK WHERE PROJECT_LINK.ROAD_NUMBER = rp.ROAD_NUMBER AND " +
-        s"PROJECT_LINK.ROAD_PART_NUMBER = rp.ROAD_PART_NUMBER AND " +
-        s"PROJECT_LINK.PROJECT_ID = rp.PROJECT_ID AND ROWNUM < 2) as pl_id " +
-        s"FROM " +
-        s"PROJECT_RESERVED_ROAD_PART rp WHERE project_id = $projectId ORDER BY road_number, road_part_number").list.map {
+  def fetchReservedRoadParts(projectId: Long, filterNotStatus: Seq[Int] = Seq.empty[Int]): Seq[ReservedRoadPart] = {
```
tvenhola: Missing unit test

tvenhola: I'll fix this too.

PR #993, opened 2017-11-10
VIITE-848 addressing roundabouts piece by piece

tvenhola:
Keep starting link even if roundabout is created piece by piece

Fix incorrect starting calibration point geometry, use original address length on transfer / unchanged etc.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/TrackSectionOrder.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/TrackSectionOrder.scala
@@ -59,6 +59,11 @@ object TrackSectionOrder {
   }
 
   def mValueRoundabout(seq: Seq[ProjectLink]): Seq[ProjectLink] = {
+    /* None = No change. Therefore for calibration points
+       None            = No change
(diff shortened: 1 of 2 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/TrackSectionOrder.scala
```diff
@@ -59,6 +59,15 @@ object TrackSectionOrder {
   }
 
   def mValueRoundabout(seq: Seq[ProjectLink]): Seq[ProjectLink] = {
+    def adjust(pl: ProjectLink, sideCode: Option[SideCode] = None, startAddrMValue: Option[Long] = None,
+               endAddrMValue: Option[Long] = None, startCalibrationPoint: Option[Option[CalibrationPoint]] = None,
```
Campanha: was Option[Option[CalibrationPoint]] intended?

tvenhola: Yes. I'll add a comment for it.
For clearing a calibration point, Some(None) is used.
For not clearing, None is used.

PR #996, opened 2017-11-13
VIITE-851 roundabouts get automatically EndOfRoad on one-click operation

tvenhola:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -268,7 +268,12 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
               logger.info("Added links recognized to be a roundabout - using roundabout addressing scheme")
               val ordered = newProjectLinks.values.toSeq.partition(_.linkId == firstLinkId)
               val created = TrackSectionOrder.mValueRoundabout(ordered._1 ++ ordered._2)
-              created
+              val endingM = created.map(_.endAddrMValue).max
```
Campanha: i think maxBy can be used here

tvenhola: Could be, probably, but won't be any cleaner.

PR #1007, opened 2017-11-16
VIITE-898

Campanha:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/ProjectLinkSplitter.scala
```diff
@@ -109,7 +109,10 @@ object ProjectLinkSplitter {
       if (isReversed)
         movedFromStart(suravageM, templateM, splitAddressM)
       else
-        movedFromEnd(suravageM, templateM, splitAddressM)
+        if(isTailConnected(suravage, templateLink))
+          movedFromEnd(suravageM, templateM, splitAddressM)
+        else
+          movedFromStart(suravageM, templateM, splitAddressM)
```
plourenco90: You could do something like:

if (isReversed && !isTailConnected(suravage, templateLink)){
moveFromStart
} else 
moveFromEnd

pedrosag: I second that motion!

PR #1010, opened 2017-11-17
Viite 850 updating a saved uusi

hestrompa:
Updating a saved Uusi link doesn't update road number, road part number or track code


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -380,7 +380,7 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
       val writableProject = projectWritable(links.projectId)
       writableProject.updateProjectLinks(links.projectId, links.linkIds.toSet,
         LinkStatus.apply(links.linkStatus), user.username,links.coordinates, links.roadNumber,
-        links.roadPartNumber, links.userDefinedEndAddressM, links.roadType, links.discontinuity, links.roadEly) match {
+        links.roadPartNumber, links.userDefinedEndAddressM, links.roadType, links.discontinuity, links.roadEly, false, links.trackCode) match {
```
oappicgi: reversed=false  would be preferred to just false

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -488,6 +488,13 @@ object ProjectDAO {
     Q.updateNA(sql).execute
   }
 
+  def updateProjectLinkNumberingAndTrack(projectId: Long, projectLinkIds: Seq[Long], newRoadNumber: Long, newRoadPart: Long, newTrackCode:Long, userName: String ): Unit ={
```
oappicgi: i am wondering if we should copy new numbering and track to projectlink seq links and use updateProjectLinksToDB to reduce amount of sql code and prefer prepared statements?

oappicgi: actually it seems that we are not using this method anywhere

PR #1013, opened 2017-11-21
Viite 873 geometry on reservation

tvenhola:
Includes generic fixes for project coordinates, default values, time stamp handling etc

### DIFF
--- a/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
+++ b/digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
@@ -93,7 +93,7 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
       val config = userProvider.getCurrentUser().configuration
       (config.east.map(_.toDouble), config.north.map(_.toDouble), config.zoom.map(_.toInt))
     }
-    StartupParameters(east.getOrElse(390000), north.getOrElse(6900000), zoom.getOrElse(2), deploy_date)
+    StartupParameters(east.getOrElse(DefaultLatitude), north.getOrElse(DefaultLongitude), zoom.getOrElse(DefaultZoomLevel), deploy_date)
   }
 
   get("/roadlinks") {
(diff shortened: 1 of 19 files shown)


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -379,10 +392,20 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
...
       val writableProject = projectWritable(links.projectId)
       writableProject.updateProjectLinks(links.projectId, links.linkIds.toSet,
-        LinkStatus.apply(links.linkStatus), user.username,links.coordinates, links.roadNumber,
+        LinkStatus.apply(links.linkStatus), user.username, links.roadNumber,
         links.roadPartNumber, links.userDefinedEndAddressM, links.roadType, links.discontinuity, links.roadEly) match {
         case Some(errorMessage) => Map("success" -> false, "errormessage" -> errorMessage)
-        case None => Map("success" -> true, "id" -> links.projectId, "publishable" -> (projectService.projectLinkPublishable(links.projectId)))
+        case None => {
+          try {
+            println("success?")
```
pedrosag: Needs to be removed.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/package.scala
```diff
@@ -73,4 +80,30 @@ package object viite {
...
+        Seq(p.x, p.y, p.z)
+      else
+        Seq(p.x, p.y)).map(toBD).mkString("[", ",","]")).mkString(",")
+  }
+
+  def toGeometry(geometryString: String): Seq[Point] = {
+    def toBD(s: String): Double = {
+      BigDecimal(s).setScale(3, BigDecimal.RoundingMode.HALF_UP).toDouble
+    }
+    val pointRegex = raw"\[[^\]]*]".r
```
pedrosag: Just a question, does this regex work usually when I try to do a single character I need to use \\[ instead of \[.
If it does, teach me your ways.

tvenhola: [a-z] is a Regex structure meaning "any character from a to z"
the caret in [^a-z] reverses the logic and it becomes "any character that is not from a to z"
For brackets one must use the backslash to escape it from being parsed as a regex rule.

PR #1014, opened 2017-11-21
Fixes to project list (VIITE-914 & VIITE-915)

hestrompa:
VIITE-914 Projects in project list are not sorted by ELY-code
VIITE-915 ELY-code and username are missing from projects rejected by Tierekisteri in project list

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
@@ -645,7 +645,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
   def getRoadAddressAllProjects(): Seq[RoadAddressProject] = {
     withDynTransaction {
-      ProjectDAO.getRoadAddressProjects().sortBy(_.ely)
+      ProjectDAO.getRoadAddressProjects()
     }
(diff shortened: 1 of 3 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -645,7 +645,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
   def getRoadAddressAllProjects(): Seq[RoadAddressProject] = {
     withDynTransaction {
-      ProjectDAO.getRoadAddressProjects()
+      ProjectDAO.getRoadAddressProjects().sortBy(_.ely)
```
tvenhola: This is the only use of getting all the projects. We should move down the sort to SQL level,

    val query =
      s"""SELECT id, state, name, created_by, created_date, start_date, modified_by, COALESCE(modified_date, created_date),
           add_info, status_info, ely, coord_x, coord_y, zoom
          FROM project $filter order by name, id """

->

    val query =
      s"""SELECT id, state, name, created_by, created_date, start_date, modified_by, COALESCE(modified_date, created_date),
           add_info, status_info, ely, coord_x, coord_y, zoom
          FROM project $filter order by ely, name, id """

PR #1018, opened 2017-11-22
Viite 870 saving project link geometry

plourenco90:


tvenhola:
Multiple tests still failing

tvenhola:
Current tests are now fixed after next commit

tvenhola:
Compiles

plourenco90:
All tests passing and fixed geometry no being draw bug

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -357,10 +367,13 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
     val user = userProvider.getCurrentUser()
     try {
       val links = parsedBody.extract[RoadAddressProjectLinksExtractor]
+      println(links)
```
oappicgi: we can probably remove this println?

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -729,7 +746,9 @@ object ProjectConverter {
     val formatter = DateTimeFormat.forPattern("dd.MM.yyyy")
     RoadAddressProject(project.id, ProjectState.apply(project.status),
       if (project.name.length > 32) project.name.substring(0, 32).trim else project.name.trim,
-      user.username, DateTime.now(), "-", formatter.parseDateTime(project.startDate), DateTime.now(), project.additionalInfo, project.roadPartList.map(toReservedRoadPart), Option(project.additionalInfo), project.projectEly, Some(ProjectCoordinates(0,0,0)))
+      user.username, DateTime.now(), "-", formatter.parseDateTime(project.startDate), DateTime.now(),
```
oappicgi: "-" would be better as user.username

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1417,7 +1381,6 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     val linkIds = projectLinks.map(_.linkId).toSet ++ expiringRoadAddresses.values.map(_.linkId).toSet
     val roadLinks = (roadLinkService.fetchVVHRoadlinks(linkIds, frozenTimeVVHAPIServiceEnabled) ++
```
oappicgi: we should probably remove these fetches?

oappicgi: to my understanding we were suppose to add project_link geometry as is, unless i am not mistaken we now fetch new geometry from vvh and then push project to roadaddress table.

tvenhola: We are able to do a ProjectLink -> RoadAddress conversion without using VVH at all. If the geometry has been updated since the change table does the conversion without problems.

PR #1020, opened 2017-11-23
VIITE-919 Tietyyppi changes when performing numbering operation

hestrompa:
Using default road type (saved in VIITE database) instead of guessing. Just guess road type when is Unknown in database

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/AddressLinkBuilder.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/AddressLinkBuilder.scala
@@ -208,7 +208,7 @@ trait AddressLinkBuilder {
           previousSegment.discontinuity
       }
 
-      Seq(RoadAddress(tempId, nextSegment.roadNumber, nextSegment.roadPartNumber, RoadType.Unknown,
+      Seq(RoadAddress(tempId, nextSegment.roadNumber, nextSegment.roadPartNumber, nextSegment.roadType,
         nextSegment.track, discontinuity, startAddrMValue,
         endAddrMValue, nextSegment.startDate, nextSegment.endDate, nextSegment.modifiedBy, nextSegment.lrmPositionId, nextSegment.linkId,
         startMValue, endMValue, nextSegment.sideCode, nextSegment.adjustedTimestamp, calibrationPoints, false, combinedGeometry,
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
@@ -420,12 +420,12 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
     * project links that are no longer reserved for the project. Reservability is check before this.
     */
   private def addLinksToProject(project: RoadAddressProject, resolution: Int): Option[String] = {
-    def toProjectLink(roadTypeMap: Map[Long, RoadType])(roadAddress: RoadAddress): ProjectLink = {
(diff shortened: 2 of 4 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -440,7 +440,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
           logger.debug(s"Reserve $reservation")
           val addressesOnPart = RoadAddressDAO.fetchByRoadPart(reservation.roadNumber, reservation.roadPartNumber, false)
           val mapping = roadLinkService.getViiteRoadLinksByLinkIdsFromVVH(addressesOnPart.map(_.linkId).toSet, false, frozenTimeVVHAPIServiceEnabled)
-            .map(rl => rl.linkId -> RoadAddressLinkBuilder.getRoadType(rl.administrativeClass, rl.linkType)).toMap
+            .map(rl => rl.linkId -> addressesOnPart.find(_.linkId == rl.linkId).head.roadType).toMap
```
tvenhola: Mapping becomes obsolete because it can be directly read from the address information. Besides finding by link id may map to incorrect road address object (there may be more than one on link) and should use mapping by road address id.(project_link.road_address_id = road_address.id in SQL)

tvenhola: line 423:    def toProjectLink(roadTypeMap: Map[Long, RoadType])(roadAddress: RoadAddress): ProjectLink = {
should drop that mapping and just do
line 428:        LinkStatus.NotHandled, roadTypeMap.getOrElse(roadAddress.linkId, RoadType.Unknown), roadAddress.linkGeomSource, GeometryUtils.geometryLength(roadAddress.geometry), roadAddress.id, roadAddress.ely, reversed = false)
->
line 428:        LinkStatus.NotHandled, roadAddress.roadType, roadAddress.linkGeomSource, GeometryUtils.geometryLength(roadAddress.geometry), roadAddress.id, roadAddress.ely, reversed = false)

tvenhola: Saw another bug in here, should instead of 
GeometryUtils.geometryLength(roadAddress.geometry)
have 
(roadAddress.endMValue - roadAddress.startMValue) because geometry is simplified in road address data.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -728,8 +728,14 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
         relatedRoadLink match {
           case None => t
           case Some(rl) =>
-            val roadType = RoadAddressLinkBuilder.getRoadType(rl.administrativeClass, rl.linkType)
-            t.copy(roadType = roadType)
+            t.roadType match {
```
tvenhola: This method is not used -> should be removed

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/AddressLinkBuilder.scala
```diff
@@ -208,7 +208,7 @@ trait AddressLinkBuilder {
           previousSegment.discontinuity
       }
 
-      Seq(RoadAddress(tempId, nextSegment.roadNumber, nextSegment.roadPartNumber, RoadType.Unknown,
+      Seq(RoadAddress(tempId, nextSegment.roadNumber, nextSegment.roadPartNumber, nextSegment.roadType,
```
tvenhola: Good find!

PR #1022, opened 2017-11-24
Viite 926, 918 drawing order in renumbering+flipping direction

oappicgi:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1022,6 +1022,7 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
             ProjectDAO.getProjectLinksByLinkId(updatedProjectLinks.head.linkId).headOption match {
               case Some(roadPartLink) =>
                 if (roadPartLink.roadNumber == newRoadNumber && roadPartLink.roadPartNumber == newRoadPartNumber)
+                  if (ProjectDAO.getProjectLinksByProjectAndLinkId(linkIds,projectId).exists(x=>x.status!=LinkStatus.Numbering)) //check if renumbering has already been done (and we are actually doing ie direction change)
```
tvenhola: Should we instead check that the renumbering is not trying to use the same number as the referenced road address object?

oappicgi: that would be cleaner way

oappicgi: i will change the logic to that

oappicgi:
didnt have change to run test on latest commit

PR #1027, opened 2017-11-28
Droth 995 bus stop expired in tiereskiteri and not in oth

MAC-CGI:
Droth 995 bus stop expired in tiereskiteri and not in oth


joanacoelho:
ok

oskarmalmstrom:
ok

PR #1028, opened 2017-11-28
returns ely to null if no roadparts have been reserved

oappicgi:
previously we  updated ely number we got from frontend request if project had no roadparts reserved -> now we force ely to null if no roadparts are reserved.

Also added ely query that stops after first row with non null ely is found so db doesn't have to get all project links

changed test to check also that removing roadparts  changes ely code to null

do not merge until  tests related to VVH change have been ignored or updated "mapping 6760" and  "Fetch suravage by boundingbox"


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -292,6 +292,13 @@ object ProjectDAO {
          """.execute
   }
 
+  def getProjectLinksFirstEly(projectId: Long): Seq[ProjectLink] = {
```
tvenhola: preferably
getElyFromProjectLinks(projectId:Long): Option[Long]

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -292,6 +292,13 @@ object ProjectDAO {
...
+     val query =
+      s"""$projectLinkQueryBase
```
tvenhola: sql"""SELECT ELY FROM PROJECT_LINK WHERE PROJECT_ID=$projectId AND ELY IS NOT NULL AND ROWNUM < 2"""...headOption

PR #1031, opened 2017-11-29
Viite 101 removing an unfinished project

hestrompa:


plourenco90:
Needs to check in what state the project has to be for the button to appear

hestrompa:
Changes applied to validate the project status before deleting info

digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -265,6 +265,23 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
     }
   }
 
+  delete("/roadlinks/roadaddress/project"){
+    val projectId = parsedBody.extract[Long]
+    try {
+      val project = ProjectDAO.getRoadAddressProjectById(projectId)
+      if (projectId != 0 && project.get.status == ProjectState.Incomplete ) {
```
tvenhola: Make a check that the project returned is not empty

PR #1032, opened 2017-11-30
Viite 935 ui road address history view

pedrosag:
UI only, no backend data yet.

### DIFF
--- a/viite-UI/src/model/RoadCollection.js
+++ b/viite-UI/src/model/RoadCollection.js
@@ -63,8 +63,11 @@
     var tmpRoadAddresses = [];
     var tmpNewRoadAddresses = [];
     var preMovedRoadAddresses = [];
+    var date = [];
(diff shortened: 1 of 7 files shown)


viite-UI/src/model/RoadCollection.js
```diff
@@ -95,32 +111,21 @@
             _.contains(selectedIds, roadLink.getId());
           });
         }).concat(getSelectedRoadLinks());
+
+        historicRoadLinks = _.filter(roadLinkGroups, function(group) {
+          groupDataSourceFilter(group, LinkSource.HistoryLinkInterface);
```
tvenhola: LinkSource relates to VVH historic links, not historic road addresses of road links -> should not be filtered at client side at all

tvenhola: Actually unrelated to the fact.

PR #1034, opened 2017-12-04
Viite 453 validation of project

pedrosag:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1493,5 +1496,69 @@ class SplittingException(s: String) extends RuntimeException {
   override def getMessage: String = s
 }
 
+class ProjectValidator {
```
tvenhola: Could this be an object class? We don't need to create multiple validator objects, we don't need to generate subclasses.

PR #1037, opened 2017-12-04
Viite 925 reserved road parts table changes

plourenco90:


digiroad2-api-viite/src/main/scala/fi/liikennevirasto/digiroad2/ViiteApi.scala
```diff
@@ -674,11 +674,13 @@ class ViiteApi(val roadLinkService: RoadLinkService, val vVHClient: VVHClient,
...
       "roadPartNumber" -> reservedRoadPart.roadPartNumber,
-      "roadPartId" -> reservedRoadPart.id,
-      "ely" -> reservedRoadPart.ely,
-      "roadLength" -> reservedRoadPart.roadLength,
-      "addressLength" -> reservedRoadPart.addressLength,
-      "discontinuity" -> reservedRoadPart.discontinuity.description,
+      "id" -> reservedRoadPart.id,
+      "currentEly" -> reservedRoadPart.ely,
+      "currentLength" -> reservedRoadPart.addressLength,
+      "discontinuity" -> reservedRoadPart.discontinuity.map(_.description),
```
Campanha: not really required but this "discontinuity" should be "currentDiscontinuity"

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -439,28 +444,82 @@ object ProjectDAO {
...
+              MAX(ra.ely) as ELY,
+              MAX(pl.ely) as ELY_NEW
+              FROM PROJECT_RESERVED_ROAD_PART rp LEFT JOIN
+              ROAD_ADDRESS ra ON (ra.road_number = rp.road_number AND ra.road_part_number = rp.road_part_number)
+              LEFT JOIN
+              PROJECT_LINK pl ON (pl.project_id = rp.project_id AND pl.road_number = rp.road_number AND pl.road_part_number = rp.road_part_number)
+              WHERE
+                rp.project_id = $projectId AND
+                RA.END_DATE IS NULL AND RA.VALID_TO IS NULL AND
+                (PL.STATUS IS NULL OR (PL.STATUS != 5 AND PL.TRACK_CODE IN (0,1)))
```
Campanha: Why are we excluding LeftSide tracks?

plourenco90: Well i believe its because the we are getting the road part length with this query and since the endaddress from last track 1 should be the same has endaress from last track2 getting one of them is enough to get the road part length

PR #1041, opened 2017-12-06
Viite 942 discontinuity checks

pedrosag:
Missing the Finnish text messages.


oappicgi:
looks good, but like @pedrosag  said messages are undefined.

PR #1049, opened 2017-12-11
Viite 398  Initial load of Road address history 19.11.2017 to devtest

plourenco90:


digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/util/DataFixture.scala
```diff
@@ -92,6 +92,14 @@ object DataFixture {
     println()
   }
 
+  def importRoadAddressesHistory(isDevDatabase: Boolean): Unit = {
```
oappicgi: i dont think isdevdatabase boolean is needed

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/AssetDataImporter.scala
```diff
@@ -328,6 +423,28 @@ class AssetDataImporter {
...
 
+  def importRoadAddressHistory(conversionDatabase: DatabaseDef, importDate: String): Unit = {
+
+    print(s"\n${DateTime.now()} - ")
+    println("Starting to import history data %s ".format(importDate))
+
+    val roadMaintainerElys = Seq(0, 1, 2, 3, 4, 8, 9, 10, 12, 14)
+
+    //Delete all the previous road history
+    OracleDatabase.withDynTransaction{
```
oappicgi: should we drop all the roadaddress and lrm table content?

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/util/DataFixture.scala
plourenco90: yes, you are right!

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/AssetDataImporter.scala
plourenco90: well, for now it is just deleting the history with "END_DATE" not null, we could delete all but we would lose the "current road" and would have to reinsert it

oappicgi: yea, thought we wanted it like that which is why i asked it on daily to be sure. Didn't run test, but code seems ok to me.

plourenco90:
It now contains VIITE-936 and VIITE-937

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/AssetDataImporter.scala
```diff
@@ -277,6 +279,129 @@ class AssetDataImporter {
...
+        (address._7, address._8, SideCode.TowardsDigitizing.value)
+      } else {
+        (address._8, address._7, SideCode.AgainstDigitizing.value)
+      }
+      val (x1, y1, x2, y2) = if (sideCode == SideCode.TowardsDigitizing.value)
+        (address._13, address._14, address._15, address._16)
+      else
+        (address._15, address._16, address._13, address._14)
+
+      print(s"insert into ROAD_ADDRESS (id, lrm_position_id, road_number, road_part_number, track_code, discontinuity, START_ADDR_M, END_ADDR_M, start_date, end_date, created_by, VALID_FROM, geometry, floating, road_type, ely, terminated) values " +
```
tvenhola: debug print

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/dao/RoadAddressDAOSpec.scala
```diff
@@ -30,6 +32,46 @@ class RoadAddressDAOSpec extends FunSuite with Matchers {
     }
   }
 
+  test("insert roadaddress duplicate info check") {
```
tvenhola: Would be better to use intercept to catch the exception and to check that error message

plourenco90: You are right, its done

PR #1052, opened 2017-12-12
VIITE-962 selection of split terminated link and form info

plourenco90:


viite-UI/src/model/SelectedProjectLink.js
```diff
@@ -22,7 +22,7 @@
 
     var orderSplitParts = function(links) {
       var splitLinks =  _.partition(links, function(link){
-        return link.roadLinkSource === LinkGeomSource.SuravageLinkInterface.value && !_.isUndefined(link.connectedLinkId);
+        return !_.isUndefined(link.connectedLinkId);
```
Campanha: i think this could lead to problems when A or B have length 0. Links will then be A + Terminated or B + Terminated -> 
var suravageB = orderedSplitParts[1]; will have terminated properties
var terminatedC = orderedSplitParts[2]; will be empty

PR #1056, opened 2017-12-13
VIITE-967 - check change information works also with roadaddress history

pedrosag:
Small adaptation to the apply_change_information_to_road_address_links in order to validate road addresses

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
@@ -280,7 +280,7 @@ object RoadAddressDAO {
     }
   }
 
-  def fetchByLinkId(linkIds: Set[Long], includeFloating: Boolean = false, includeHistory: Boolean = true, includeValid: Boolean = false): List[RoadAddress] = {
+  def fetchByLinkId(linkIds: Set[Long], includeFloating: Boolean = false, includeHistory: Boolean = true, includeTerminated: Boolean = true): List[RoadAddress] = {
     if (linkIds.size > 1000) {
       return fetchByLinkIdMassQuery(linkIds, includeFloating, includeHistory)
     }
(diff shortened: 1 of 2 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -298,6 +298,12 @@ object RoadAddressDAO {
     else
       ""
 
+    val valid = if(includeValid) {
+      "AND (ra.end_date is null OR ra.end_data >= sysdate) AND ra.terminated = 0 "
```
tvenhola: Should not include a date check: all road address history should be moved with vvh change table to the new link. Instead only termination should be checked.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -280,7 +280,7 @@ object RoadAddressDAO {
     }
   }
 
-  def fetchByLinkId(linkIds: Set[Long], includeFloating: Boolean = false, includeHistory: Boolean = true): List[RoadAddress] = {
+  def fetchByLinkId(linkIds: Set[Long], includeFloating: Boolean = false, includeHistory: Boolean = true, includeValid: Boolean = false): List[RoadAddress] = {
```
tvenhola: Should not be called "valid" as it falsely points to valid_from and valid_to fields.
includeTerminated: Boolean = true

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
tvenhola: use includeTerminated flag and change to 
if (!includeTerminated) {
  "AND ra.terminated = 0 "
...

Campanha: "ra.end_data" instead "ra.end_date" should have done problems tho

PR #1066, opened 2017-12-19
Viite 968 ui boundingbox fetch

pedrosag:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -911,6 +911,16 @@ object RoadAddressDAO {
     } else {
       s""" where ra.id in ($idString)"""
     }
+    val terminatedFilter = if(includeTerminated) {
```
tvenhola: Condition should be inverted here

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -470,8 +472,12 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
 
   def mergeRoadAddress(data: RoadAddressMerge): Unit = {
-    withDynTransaction {
-      mergeRoadAddressInTX(data)
+    try {
+      withDynTransaction {
+        mergeRoadAddressInTX(data)
+      }
+    } catch {
+      case ex: InvalidAddressDataException => logger.info("Duplicate merging(s) found, skipped")
```
tvenhola: Add a generic exception handler with logger.error(..., ex)

PR #1068, opened 2017-12-20
VIITE-984 first selection fix

plourenco90:


plourenco90:
the local one was removed, so i don't think it can be an issue

pedrosag:
@plourenco90 As I said, just careful :3 but I aprove of this

PR #1069, opened 2017-12-20
VIITE-983 - Jakamalla JATKUU-koodin muutos on tallennettava vain A-tai B-osalle

pedrosag:
Added Discontinuity adjustment method to the pre-split links.


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -523,6 +523,20 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
 
   def preSplitSuravageLinkInTX(linkId: Long, username: String,
                                splitOptions: SplitOptions): (Option[Seq[ProjectLink]], Option[String], Option[(Point, Vector3d)]) = {
+    //Discontinuity of splits should be the given only for the part with the biggest M Address Value
```
tvenhola: I think this would better fit ProjectLinkSplitter.scala where all the other adjustment logic resides too. For example     
def switchDigitization(splits: (ProjectLink, ProjectLink, ProjectLink)) = ...

pedrosag: I agree, switching.

PR #1071, opened 2017-12-21
VIITE-991 - Selection on the map is too large when splitting

pedrosag:
Changed the linkIds that are loaded on the SelectedProjectLink.
Decreased the zIndex of the transferred Link (dark yellow).

### DIFF
--- a/viite-UI/src/view/ProjectLinkLayer.js
+++ b/viite-UI/src/view/ProjectLinkLayer.js
@@ -631,7 +631,7 @@
         });
         var style = new ol.style.Style({
           stroke: new ol.style.Stroke({color: '#c6c00f', width: 13, lineCap: 'round'}),
-          zIndex: 11
+          zIndex: -1
         });
         terminatedFeature.setStyle(style);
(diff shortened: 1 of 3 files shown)


viite-UI/src/view/ProjectLinkLayer.js
```diff
@@ -631,7 +631,7 @@
         });
         var style = new ol.style.Style({
           stroke: new ol.style.Stroke({color: '#c6c00f', width: 13, lineCap: 'round'}),
-          zIndex: 11
+          zIndex: -1
```
Campanha: both terminated and its marker "C" have zIndex = 11. Dunno if one change then the other one should also change. 
Still this change seems visually good since terminated marker will be visible (e.g. in top of suravage links)

oappicgi:
Otherwise ok, but occasionally yellow template link is over suravage link. Would be nice if that could be fixed I couldn't find reason for it. Mainly because i couldn't figure out logic when they are in wrong order

PR #1078, opened 2017-12-27
DROTH-1072 Speed Limit gaps fix

joanacoelho:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/util/TierekisteriImporterOperations.scala
```diff
@@ -402,80 +404,45 @@ class SpeedLimitsTierekisteriImporter extends LinearAssetTierekisteriImporterOpe
...
+    def getDefaultTrTrafficSign(addressSection: AddressSection): Option[TierekisteriAssetData] ={
+      Some(TierekisteriTrafficSignData(addressSection.roadNumber, addressSection.roadPartNumber, addressSection.roadPartNumber, addressSection.track, addressSection.startAddressMValue, addressSection.startAddressMValue, roadSide, TRTrafficSignType.EndUrbanArea, defaultSpeedLimit.toString))
     }
 
-    //Middle Part of The Section Splited
-    if ((trUAAsset.startAddressMValue >= addressSection.startAddressMValue) && (trUAAsset.endAddressMValue <= addressSection.endAddressMValue.head)) {
-      val newMiddleAddressSection = addressSection.copy(startAddressMValue = trUAAsset.startAddressMValue, endAddressMValue = Some(trUAAsset.endAddressMValue))
-      createSectionPart(roadAddress, newMiddleAddressSection, TRTrafficSignType.UrbanArea, urbanAreaSpeedLimit.toString)
-    }
+    val trUrbanAssets = trUrbanAreaAssets.filter(_.startRoadPartNumber == addressSection.roadPartNumber)
```
leirasf: Here we are just take care of the start road part number, the urban area is a linear asset like, so we should taking care of the start and the end. The idea of reusing the existing code for the traffic signs is good we just need to convert in the urban area linear asset in two point urban area, start and the end.

leirasf:
Beside the previous comment, it's ok to me

leirasf:
OK

PR #1079, opened 2017-12-27
Viite 977 drawing generalization on small scales

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -849,6 +856,45 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
+  def fetchProjectRoadLinksLinearGeometry(projectId: Long, boundingRectangle: BoundingRectangle, roadNumberLimits: Seq[(Int, Int)], municipalities: Set[Int],
+                            everything: Boolean = false, publicRoads: Boolean = false): Seq[ProjectAddressLink] = {
+
+    val fetchRoadAddressesByBoundingBoxF = Future(withDynTransaction {
+      val (floating, addresses) = RoadAddressDAO.fetchRoadAddressesByBoundingBox(boundingRectangle, fetchOnlyFloating = false,
+        roadNumberLimits = roadNumberLimits).partition(_.floating)
+      (floating.groupBy(_.linkId), addresses.groupBy(_.linkId))
+    })
+
+    val fetchProjectLinksF = Future(withDynSession(ProjectDAO.getProjectLinks(projectId).groupBy(_.linkId)))
```
oappicgi: just wondering if there is any reason to use future if we just immediately wait for it to complete?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressLinkBuilder.scala
```diff
@@ -38,6 +38,24 @@ object RoadAddressLinkBuilder extends AddressLinkBuilder {
 
   }
 
+  def buildSimpleLink(roadAddress: RoadAddress): RoadAddressLink = {
```
oappicgi: maybe one simple test for this method?

tvenhola: Might be too much overhead as it's only a really simple mapping.

PR #1082, opened 2017-12-29
Viite 958 discontinuity checks

tvenhola:
Discontinuity checks for project road parts.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
@@ -151,20 +151,14 @@ object ProjectValidator {
     project.reservedParts.filter(rrp => rrp.addressLength.nonEmpty && rrp.newLength.getOrElse(0L) == 0L &&
       rrp.discontinuity.contains(EndOfRoad))
 
-      // There is no part after this part in project
-      .filterNot(rrp => project.reservedParts.exists(np => np.roadNumber == rrp.roadNumber &&
-      np.roadPartNumber > rrp.roadPartNumber &&  np.newLength.getOrElse(0L) > 0L))
-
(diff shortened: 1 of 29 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectValidator.scala
```diff
@@ -0,0 +1,377 @@
...
+
+  object ValidationError {
+    val values = Set(MinorDiscontinuityFound, MajorDiscontinuityFound, InsufficientTrackCoverage, DiscontinuousAddressScheme,
+      SharedLinkIdsExist, NoContinuityCodesAtEnd, UnsuccessfulRecalculation, MissingEndOfRoad)
+    //Viite-942
+    case object MissingEndOfRoad extends ValidationError {def value = 0
+      def message = MissingEndOfRoadMessage
+      def notification = false}
+    //Viite-453
+    //There must be a minor discontinuity if the jump is longer than 0.1 m (10 cm) between road links
```
oappicgi: maybe instead of values we should have variable location where value can be change so comment doesn't have to be maintained?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/ProjectDAO.scala
```diff
@@ -473,6 +472,7 @@ object ProjectDAO {
   }
 
   def fetchReservedRoadParts(projectId: Long, filterNotStatus: Seq[Int] = Seq.empty[Int]): Seq[ReservedRoadPart] = {
+    // TODO: Check this filter: is it required to filter all road parts where ANY of the project links status is in filterNotStatus?
```
Campanha: the *filterNotStatus* seems to be not be passing through methods.  The filter could probably be removed.

PR #1083, opened 2017-12-29
Viite 473 unchanged validation with others operations in chain

hestrompa:


digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/ProjectServiceSpec.scala
```diff
@@ -1216,4 +1217,23 @@ class ProjectServiceSpec extends FunSuite with Matchers with BeforeAndAfter {
       validationErrors.head.validationError.value should be (0)
     }
   }
+
+  test("validator should return invalid unchanged links error"){
```
tvenhola: Should be moved to ProjectValidatorSpec class

PR #1084, opened 2017-12-29
Droth 1109 winter roads need to be filtered out in vvh api

joanacoelho:


oskarmalmstrom:
ok

leirasf:
We need to be careful with that because VIITE uses the same service with the same enrich methods, so if we want to have some specific for OTH we should extend the roadlinkservice.

oskarmalmstrom:
ok again

leirasf:
Ok

PR #1085, opened 2018-01-02
VIITE-686 + VIITE-980 add viite ui tests in jenkins run dr2 build and test fixes

plourenco90:


gruntfile.js
```diff
@@ -179,7 +179,7 @@ module.exports = function(grunt) {
           {
             context: '/test/components',
             host: 'localhost',
-            port: '9001',
+            port: '9003',
```
oappicgi: why do we change port?

plourenco90: Because we can't run two web servers on same port.

PR #1096, opened 2018-01-11
Viite 1035 send button enabler validation

oappicgi:
Send button is enabled only if validations pass and project change table is open. Changed publishable method to check validations and renamed old method to it's current use of checking if all links have been processed. This way we are not doing same thing twice


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1096,8 +1096,16 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
...
+  /**
+    * method to check if project is publishable. add filters for cases we do not want to prevent sending
+    * @param projectId project-id
+    * @return if project contains any notifications preventing sending
+    */
+  def isProjectPublishable(projectId: Long): Boolean = {
+    validateProjectById(projectId).isEmpty
+  }
+
+  def allLinksHandeled(projectId: Long): Boolean ={ //some tests want to know if all projectlinks have been handeled. to remove this test need to be updated to check if termination is correctly applied etc best done after all validations have been implemented
```
Campanha: there is a typo in Handeled -> Handled

PR #1099, opened 2018-01-11
Viite 1012 floating road address in devtest

pedrosag:
Changes summary:

- Added the additional filters to the Mass Query (they were not existing).
- Changed all the truncate to it's 2D versions.
- Added a special check that if a road address geometry is found to be "moved" when in comparison to the original VVH roadlink geometry, we check if said road address was a result of the automatic merging process and if said road address overlaps the original VVH geometry; if it true to both then the road is not floating.
- Increased the amount and details of console prints on the console for the find_road_address batch process.

### DIFF
--- a/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/FloatingChecker.scala
+++ b/digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/FloatingChecker.scala
@@ -16,7 +16,7 @@ class FloatingChecker(roadLinkService: RoadLinkService) {
     def outsideOfGeometry(ra: RoadAddress, roadLinks: Seq[RoadLinkLike]): Boolean = {
       !roadLinks.exists(rl => GeometryUtils.geometryLength(rl.geometry) > ra.startMValue) ||
         !roadLinks.exists(rl => GeometryUtils.areAdjacent(
-          GeometryUtils.truncateGeometry3D(rl.geometry, ra.startMValue, ra.endMValue),
+          GeometryUtils.truncateGeometry2D(rl.geometry, ra.startMValue, ra.endMValue),
           ra.geometry))
     }
 
@@ -31,6 +31,12 @@ class FloatingChecker(roadLinkService: RoadLinkService) {
         val rl = roadLinks(ra.linkId).head
         val len = GeometryUtils.geometryLength(rl.geometry)
         println(s"${pretty(ra)} moved to floating, outside of road link geometry (link is $len m)")
+        //TODO: Testing - No update
+        println(s"Road Address: ${ra.geometry}")
+        println(s"Road Address ID: ${ra.id}")
+        println(s"Road Address Created/Modified: ${ra.modifiedBy}")
+        println(s"Road Link: ${rl.geometry}")
+        println("---------")
       }
     )
     val floatings = checkGeometryChangeOfSegments(roadAddressList, roadLinks)
@@ -60,7 +66,7 @@ class FloatingChecker(roadLinkService: RoadLinkService) {
   }
 
   /**
-    * Check if road address geometry is moved by road link geometry change at leas MaxMoveDistanceBeforeFloating
+    * Check if road address geometry is moved by road link geometry change at least MaxMoveDistanceBeforeFloating
     * meters. Because road address geometry isn't directed check for fit either way. Public for testing.
     * @param roadLink Road link for road address list
     * @param roadAddresses Sequence of road addresses to check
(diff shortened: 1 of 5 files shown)


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/FloatingChecker.scala
```diff
@@ -16,8 +16,8 @@ class FloatingChecker(roadLinkService: RoadLinkService) {
     def outsideOfGeometry(ra: RoadAddress, roadLinks: Seq[RoadLinkLike]): Boolean = {
       !roadLinks.exists(rl => GeometryUtils.geometryLength(rl.geometry) > ra.startMValue) ||
         !roadLinks.exists(rl => GeometryUtils.areAdjacent(
-          GeometryUtils.truncateGeometry3D(rl.geometry, ra.startMValue, ra.endMValue),
-          ra.geometry))
+          GeometryUtils.truncateGeometry2D(rl.geometry, ra.startMValue, ra.endMValue),
```
oappicgi: i got really bad feeling about 3D->2D change.  Is there any other way to fix this issue?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/FloatingChecker.scala
```diff
@@ -60,31 +65,50 @@ class FloatingChecker(roadLinkService: RoadLinkService) {
...
+        GeometryUtils.truncateGeometry2D(roadLink.geometry, ra.startMValue, ra.endMValue),
         ra.geometry) &&
         GeometryUtils.geometryMoved(MaxMoveDistanceBeforeFloating)(
-          GeometryUtils.truncateGeometry3D(roadLink.geometry, ra.startMValue, ra.endMValue),
+          GeometryUtils.truncateGeometry2D(roadLink.geometry, ra.startMValue, ra.endMValue),
           ra.geometry.reverse) // Road Address geometry isn't necessarily directed: start and end may not be aligned by side code
     }
-    ) || Math.abs(roadAddresses.maxBy(_.endMValue).endMValue - GeometryUtils.geometryLength(roadLink.geometry)) > MaxMoveDistanceBeforeFloating
+    )
+    if(movedAddresses.size != 0) {
```
hestrompa: Use Scala functionalities in this kind of validations. moveAddresses.isNonEmpty could do the work, or even better, avoiding the "if else" logic and use "match" to make the code more readable

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/FloatingChecker.scala
```diff
@@ -16,8 +16,8 @@ class FloatingChecker(roadLinkService: RoadLinkService) {
     def outsideOfGeometry(ra: RoadAddress, roadLinks: Seq[RoadLinkLike]): Boolean = {
       !roadLinks.exists(rl => GeometryUtils.geometryLength(rl.geometry) > ra.startMValue) ||
         !roadLinks.exists(rl => GeometryUtils.areAdjacent(
-          GeometryUtils.truncateGeometry3D(rl.geometry, ra.startMValue, ra.endMValue),
-          ra.geometry))
+          GeometryUtils.truncateGeometry2D(rl.geometry, ra.startMValue, ra.endMValue),
+          ra.geometry.map(_.copy(z = 0.0))))
```
Campanha: truncateGeometry2D automatically assumes z = 0.0 if z empty.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/process/FloatingChecker.scala
oappicgi: is the problem in data (missing Z) or why do we have to convert 3D->2D?

oappicgi: i mean if that is the case, we probably want to talk with JanneG about that. FTA might want to know if there is problem with the data

Campanha: i think Guilherme did this decause of the adjacency check. And since its comparing roadLinks[with N Point[x,y, and possibly z] with roadAddress endPoints [with only  N Point[x,y]) we wanted to ensure that z is always 0.0 for roadlink Points and z always 0.0 for roadAddress Points.
Still the *GeometryUtils.distance2DTo* compares adjacency by x and y only so in my opinion both the *truncateGeometry2D* and *truncateGeometry3D* could be used and also the *ra.geometry.map(_.copy(z = 0.0))* could be removed

pedrosag: As @Campanha pointed out it is exactly that, also adding to it, the geometry we save on our own Road_Address table is already 2D truncated, therefore it was a way to keep the playing field in the same level.

PR #1102, opened 2018-01-12
Viite 1016 viite accepts track code 99 and freezes

hestrompa:
This branch contains:
- Addition of dropdown to choose track_code to project_link
- Update to project errors list after "uusi" and "revert" operations.


viite-UI/src/model/ProjectCollection.js
```diff
@@ -254,6 +254,12 @@
         coordinates:coordinates
       };
 
+      if(dataJson.trackCode === 99){
```
Campanha: we could take this Track values into LinkValues.js

PR #1105, opened 2018-01-16
Viite 1034 api should use only the latest verified version of road addresses

plourenco90:


oappicgi:
think this task has been misunderstood. PO uses "viiteapi" name for outside sources querying viite for publishable roadnetwork, which should always deliver latest version.  what we ask  in UI from backend should be left as it is.

plourenco90:
The only change that was made, was in the query that is used in the VIITE integration api to fetch the road network. It won't interfere with the UI.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -415,6 +444,30 @@ object RoadAddressDAO {
...
+        ra.discontinuity, ra.start_addr_m, ra.end_addr_m, ra.lrm_position_id, pos.link_id, pos.start_measure, pos.end_measure,
+        pos.side_code, pos.adjusted_timestamp,
+        ra.start_date, ra.end_date, ra.created_by, ra.valid_from, ra.CALIBRATION_POINTS, ra.floating, t.X, t.Y, t2.X, t2.Y, link_source, ra.ely, ra.terminated,
+        net.id as road_version, net.created as version_date
+        from road_address ra cross join
+        TABLE(SDO_UTIL.GETVERTICES(ra.geometry)) t cross join
+        TABLE(SDO_UTIL.GETVERTICES(ra.geometry)) t2
+        join lrm_position pos on ra.lrm_position_id = pos.id
+        join $idTableName i on i.id = pos.link_id
+        join published_road_network net on net.id = (select network_id from published_road_address where ra.id = road_address_id)
```
oappicgi: since we want latest roadnetwork that has been published wouldn't that be  to join with max net.id?

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -415,6 +444,30 @@ object RoadAddressDAO {
...
+        where t.id < t2.id and
+          (valid_from is null or valid_from <= sysdate) and
```
oappicgi: valid_from cannot be null

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/dao/RoadAddressDAO.scala
```diff
@@ -342,12 +342,41 @@ object RoadAddressDAO {
...
+        select ra.id, ra.road_number, ra.road_part_number, ra.road_type, ra.track_code,
+        ra.discontinuity, ra.start_addr_m, ra.end_addr_m, ra.lrm_position_id, pos.link_id, pos.start_measure, pos.end_measure,
+        pos.side_code, pos.adjusted_timestamp,
+        ra.start_date, ra.end_date, ra.created_by, ra.valid_from, ra.CALIBRATION_POINTS, ra.floating, t.X, t.Y, t2.X, t2.Y, link_source, ra.ely, ra.terminated,
+        net.id as road_version, net.created as version_date
+        from road_address ra cross join
+        TABLE(SDO_UTIL.GETVERTICES(ra.geometry)) t cross join
+        TABLE(SDO_UTIL.GETVERTICES(ra.geometry)) t2
+        join lrm_position pos on ra.lrm_position_id = pos.id
+        join published_road_network net on net.id = (select network_id from published_road_address where ra.id = road_address_id)
```
oappicgi: max here as well like it is in other new quary?

PR #1106, opened 2018-01-16
Viite 1023 Suravage links included in the initial road address download

skosonen:
Change the import road address batch to get also the suravage VVH-links.
Change the update road address geometry batch to handle the suravage links as well.

### DIFF
--- a/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/VVHClient.scala
+++ b/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/VVHClient.scala
@@ -1170,6 +1170,10 @@ class VVHSuravageClient(vvhRestApiEndPoint: String) extends VVHRoadLinkClient(vv
   }
 
   def fetchSuravageByLinkIdsF(linkIds: Set[Long] = Set()): Future[Seq[VVHRoadlink]] = {
-    Future(queryByLinkIds(linkIds, None, true, extractRoadLinkFeature, withLinkIdFilter))
+    Future(fetchSuravageByLinkIds(linkIds))
+  }
+
+  def fetchSuravageByLinkIds(linkIds: Set[Long] = Set()): Seq[VVHRoadlink] = {
+    queryByLinkIds(linkIds, None, true, extractRoadLinkFeature, withLinkIdFilter)
   }
 }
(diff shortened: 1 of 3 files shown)


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/VVHClient.scala
```diff
@@ -1170,6 +1170,10 @@ class VVHSuravageClient(vvhRestApiEndPoint: String) extends VVHRoadLinkClient(vv
   }
 
   def fetchSuravageByLinkIdsF(linkIds: Set[Long] = Set()): Future[Seq[VVHRoadlink]] = {
-    Future(queryByLinkIds(linkIds, None, true, extractRoadLinkFeature, withLinkIdFilter))
+    Future(fetchSuravageByLinkIds(linkIds))
+  }
+
+  def fetchSuravageByLinkIds(linkIds: Set[Long] = Set()): Seq[VVHRoadlink] = {
+    queryByLinkIds(linkIds, None, true, extractRoadLinkFeature, withLinkIdFilter)
```
Campanha: I've think better about this. I realized that in lrmListWithLinkSources we are using the linkSource from roadLinks which are all being build by extractRoadLinkFeature() when queryByLinkIds() that gives LinkGeomSource.NormalLinkInterface by default.

Should we modify the linkSource within each diferent vvhEndPoint class? 
Making something like 
*queryByLinkIds(linkIds, None, true, extractRoadLinkFeature, withLinkIdFilter).map(_.copy(linkSource = linkGeomSource))*

Campanha: if we use this approach we need to probably define a new overriden fetchByLinkIds also for the VVHComplementaryClient

oappicgi: linkgeomsource should ba saved in VVHClient when we fetch roadlinks

oappicgi: for example for suravage it is set in (VVHClient) row 1161 and for complimentary 1025.

skosonen: In VVHClient.scala:633 we set linkGeomSource, which is defined in each VVHRoadLinkClient implementation. For example VVHSuravageClient has LinkGeomSource.SuravageLinkInterface.

PR #1107, opened 2018-01-16
Viite 472 checking part length filled manually

Campanha:
1. Added new method for geometry length calculation.
2. Throw warning if new endAddrM value for roadPart is less or bigger than 5% of the real road part length (by geometry)
3. Removed ELY code change validation between road parts.
4. Fixed GenericConfirmPopup for content message box in "Yes" or "No" buttons.


viite-UI/src/utils/GeometryUtils.js
```diff
@@ -220,26 +215,43 @@
...
+      return _.reduce(geometry , function (length, point, index, array) {
+        if (index < array.length - 1) {
+          return length + distanceBetweenPoints(point, array[index+1]);
+        }
+        else {
+          return length;
+        }
+      }, 0.0);
+    }
+  };
```
leirasf: You can remove redundante code, something like:
```
return _.reduce(geometry , function (length, point, index, array) {
	if (index < array.length - 1) 
		return length + distanceBetweenPoints(point, array[index+1]);
	return length;
}, 0.0);
```

viite-UI/src/model/ProjectCollection.js
```diff
@@ -230,6 +231,21 @@
     };
 
     this.saveProjectLinks = function(changedLinks, statusCode) {
+      var thisone = this;
+      var validUserGivenAddrMValues = function(linkId, userEndAddr){
+        if(_.isUndefined(userEndAddr))
+          return false;
+          var roadPartIds = self.getMultiSelectIds(linkId);
+          var roadPartLinks = self.getByLinkId(roadPartIds);
+          var sortedGeometries = _.map(_.sortBy((roadPartLinks), function(link){return link.getData().endAddressM;}), function(roadPart){
```
leirasf: I think you don't need to sort the roadPartLinks, just sum the length of each ones.

PR #1116, opened 2018-01-19
Viite 1041 view mode road addresses on suravage links

pedrosag:


viite-UI/src/view/Styler.js
```diff
@@ -33,8 +33,8 @@
      */
     var opacityMultiplier=1;
 
-    var generateStrokeColor = function (roadClass, anomaly, constructionType, roadLinkType, gapTransfering, roadLinkSource) {
-      if(roadLinkSource === LINKSOURCE_SURAVAGE) {
+    var generateStrokeColor = function (roadClass, anomaly, constructionType, roadLinkType, gapTransfering, roadLinkSource, roadId) {
+      if(roadLinkSource === LINKSOURCE_SURAVAGE && roadId === 0) {
```
oappicgi: && nonsavedroad instead of roadid=0000 would be better, so it is easier to know why we check roadid===0

PR #1122, opened 2018-01-26
Droth 1080 - Visualization of TR speed limits

marquesfm:


UI/src/view/point_asset/trafficSignReadOnlyLayer.js
```diff
@@ -66,8 +66,8 @@
         me.removeLayerFeatures();
     };
 
-    eventbus.on(parentLayerName + ':hideReadOnlyTrafficSigns', hideReadOnlyTrafficSigns);
-    eventbus.on(parentLayerName + ':showReadOnlyTrafficSigns', showReadOnlyTrafficSigns);
+    eventbus.on(parentLayerName + '-readOnlyTrafficSigns:show', hideReadOnlyTrafficSigns);
+    eventbus.on(parentLayerName + '-readOnlyTrafficSigns:hide', showReadOnlyTrafficSigns);
```
marquesfm: The events are switched, to hide the ReadOnly TrafficSigns we are using the event to show them and to show we are using the event to hide.

UI/src/utils/styleRule.js
```diff
@@ -110,6 +110,16 @@
             return this;
         };
 
+      this.isBetween = function(values){
+        var expression = expressionFn[expressionFn.length-1];
```
joanacoelho: Probably could be checked if values is an array and if it has max/min lenght = 2

UI/src/view/linear_asset/linearAssetLabel.js
```diff
@@ -62,7 +62,68 @@
...
+        if(config)
+          return config;
+
+        return {image: 'images/warningLabel.png', scale: 1};
+      };
+
+      var textStyle = function(value) {
+        if (!me.isValidValue(value))
+          return '';
+        return "" + value;
```
joanacoelho: To be coherent , change to :  **return '' + value;**

UI/src/view/linear_asset/trSpeedLimitStyle.js
```diff
@@ -0,0 +1,116 @@
+(function(root) {
+  root.TRSpeedLimitStyle = function() {
+    AssetStyle.call(this);
```
joanacoelho: We can start to think on a solution to speedLimitstyles :)

UI/src/view/navigationpanel/pointAssetBox.js
```diff
@@ -0,0 +1,115 @@
...
+    this.title = function () {
+      return assetConfig.title;
+    };
+
+    this.layerName = function () {
+      return assetConfig.layerName;
+    };
+
+    this.elements = function (){
+      return { expanded: $([
```
joanacoelho: This should be checked for point assets! The div closes in the different place as all the other assets.

UI/src/view/navigationpanel/speedLimitBox.js
```diff
@@ -0,0 +1,109 @@
+(function(root) {
+  root.SpeedLimitBox = function (selectedSpeedLimit) {
+    ActionPanelBox.call(this);
```
joanacoelho: This one also can extend from the LinearAssetBox

oskarmalmstrom:
ok

UI/src/utils/styleRule.js
```diff
@@ -110,6 +110,16 @@
...
+        if(!expression || expression.compare)
+          throw 'You must have on of the following functions ["where", "and", "or"] before use the "isBetween".';
+        expression.compare = function(propertyValue){
+            return values[0] <= propertyValue  && propertyValue < values[1];
```
leirasf: Should check the length of the values and throw a exception if it's diferente then 2.

UI/src/view/navigationpanel/speedLimitBox.js
joanacoelho: After analysing it, it's not needed to extend from LinearAssetBox, since we have to redefine everything

UI/src/utils/styleRule.js
joanacoelho: Done

UI/src/view/navigationpanel/pointAssetBox.js
joanacoelho: This div is used to separate the checkbox from the labeling

PR #1123, opened 2018-01-30
Viite 1092 performance improvement of viite api

plourenco90:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -613,22 +613,10 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
-      else tempRoadLinks
-    }
-    val complimentaryLinks = {
-      val tempComplimentary = roadLinkService.getComplementaryRoadLinksFromVVH(municipality)
-      if (tempComplimentary == null)
-        Seq.empty[RoadLink]
-      else tempComplimentary
-    }
+
+    val (roadLinksWithComplementary, _) = roadLinkService.getRoadLinksWithComplementaryAndChangesFromVVH(municipality)
```
oappicgi: are we loosing frozentime boolean effect here though?

plourenco90: yes, i will put a clause to use the frozen method if flag is active, but it will be without the cache.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -613,22 +613,27 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
...
-      else tempRoadLinks
-    }
-    val complimentaryLinks = {
-      val tempComplimentary = roadLinkService.getComplementaryRoadLinksFromVVH(municipality)
-      if (tempComplimentary == null)
-        Seq.empty[RoadLink]
-      else tempComplimentary
-    }
+
+    val (roadLinksWithComplementary, _) =
```
Campanha: do we really need two val definition here?

pedrosag: I do believe so if you check bellow on the allRoadLinks (line 637) it is using the roadLinksWithComplementary value

Campanha: yes we need that value, the left one from the definition. the , _ can be ignored if we change
"(roadLinks ++ complimentaryLinks, Seq())" to (roadLinks ++ complimentaryLinks)
and roadLinkService.getRoadLinksWithComplementaryAndChangesFromVVH(municipality) to roadLinkService.getRoadLinksWithComplementaryAndChangesFromVVH(municipality)._1

plourenco90: It is using the two val because the method that gets the cached road links also gets the changed. And so that it also works with the frozen i just added the Seq(). Since the frozen if statement part will be removed soon.

PR #1132, opened 2018-02-02
Viite 1053 jatkuvuus code validation ely border

pedrosag:
Added new validation for the ELY code when in relation with:

- **A.** The discontinuity of the project links

- **B.** The crossing of an address through a ELY border.

There are 2 checks:

If Discontinuity-code is "ELY:n raja" (3), the ELY-code of the next road part should be checked.
If ELY-code of the next road part is different there should be no validation error but if ELY-code is same: ERROR project validator will raise an error.

If Discontinuity-code is something else than ELY:n raja (3) and if next road part ELY-code is different then project validator will raise an error.

Also made a small adjustment on the front end, in the case that we have lots of linkIds on the validation error message then it would break the form, adding a scroll bar on the form alone, now it will wrap on the form.


oappicgi:
tests need to be updated, working on them atm

PR #1135, opened 2018-02-06
Viite 1106 revert lrm position merge

leirasf:


oappicgi:
just wondering if we should somehow mark places where we copy commonhistory id. I think previous idea was that we mark them as new and check later where we can copy and where we need new common_history id (basically inside project we probably get new id:s)

Campanha:
At the beginning we where in fact giving just new common history id everywhere since we didnt know whether it should or not get the old commonHistoryId. Now we have some rules where we should apply one New CommonHistoryId for segments. (e.g.: link extends/shorts through user or vvh change info interface or track code or roadtype changes through user). Other situations should keep old commonHistoryIds.

skosonen:
I added TODO-comments in those places where we need to check later how to handle those common_history_id:s.

PR #1136, opened 2018-02-06
VIITE-1110 VIITE API contains road address history

pedrosag:
Included in fetchByLinkIdToApi and the mass query a filter to not fetch historic links.


pedrosag:
Thanks @skosonen :P

PR #1137, opened 2018-02-07
Viite 1064 fixes in split with multiples links in linkid

hestrompa:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -536,8 +535,14 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
       val rightTop = Point(x._2, endPoints._2.y)
       val leftBottom = Point(x._1, endPoints._1.y)
       val projectLinks = getProjectLinksInBoundingBox(BoundingRectangle(leftBottom, rightTop), projectId)
+      val roadLink = roadLinkService.getRoadLinkByLinkIdFromVVH(projectLinks.head.linkId, false).headOption
+      if(roadLink.isEmpty){
+        (None, Some(ErrorSuravageLinkNotFound), None)
+      }
+
       val (projectLinksConnected, projectLinksDisconnected) = projectLinks.partition(l =>
```
skosonen: projectLinksDisconnected is not used any more according to IDEA

oappicgi:
tried splitting with 6/401   where pointer is and got contact admin message.
coordinates: 6948398,642599

sniping

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/ProjectLinkSplitter.scala
```diff
@@ -135,35 +135,42 @@ object ProjectLinkSplitter {
   }
 
   def split(suravage: ProjectLink, templateLink: ProjectLink, splitOptions: SplitOptions): Seq[ProjectLink] =
-    split(suravage, templateLink, Seq(), splitOptions).toSeqWithAllTerminated
+    split(suravage, templateLink, splitOptions)
```
skosonen: Here we have an infinite recursion, which can be seen when running ProjectLinkSplitterSpec: "Split aligning project links".

Campanha: I've added one fix for the older split test, and also added one new when there are 2 terminated links with same linkId. @Hugo can check the last one tho so see it needs some more rule/validation

PR #1139, opened 2018-02-07
DROTH-1164 Missing value fields in Kalpa API

oskarmalmstrom:
-added numeral and textual value services to IntegrationApi


joanacoelho:
Ok

joanacoelho:
ok

PR #1140, opened 2018-02-07
VIITE-1089 Add common history calibration points to the initial import

leirasf:


skosonen:
Test AssetDataImporterSpec "Should have calibration points where common history changes" was created to make sure that this works as wanted, but now it fails. (It is ignored, because database rollback didn't seem to work properly and this implementation was not yet done)

PR #1141, opened 2018-02-07
Viite 1107 update unit tests due to new common history id from changeinfo interface

Campanha:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1406,11 +1406,13 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
             // Transferred part, original values
             roadAddress.copy(id = NewRoadAddress, startAddrMValue = startAddr, endAddrMValue = endAddr,
               endDate = Some(project.startDate), modifiedBy = Some(project.createdBy), startMValue = startM, endMValue = endM),
+//              endDate = Some(project.startDate), modifiedBy = Some(project.createdBy), startMValue = startM, endMValue = endM, commonHistoryId = commonHistoryIdDecider(pl, roadAddress)),
```
plourenco90: Should be removed

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/ProjectService.scala
```diff
@@ -1487,6 +1488,13 @@ class ProjectService(roadAddressService: RoadAddressService, roadLinkService: Ro
         setEndDate(roadAddresses(pl.roadAddressId), pl, None))
   }
 
+//  def commonHistoryIdDecider(pl: ProjectLink, source: RoadAddress): Long = {
```
plourenco90: Will we need it in future?

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/RoadAddressServiceSpec.scala
```diff
@@ -794,7 +794,7 @@ class RoadAddressServiceSpec extends FunSuite with Matchers{
         createRoadAddressLink(Sequences.nextViitePrimaryKeySeqValue, 5622953, o5622953Geom, 2825, 3, 0, 279, 321, SideCode.TowardsDigitizing, Anomaly.None, false, true) // end calibration point for testing
       )
 
-      val addresses = oldAddressLinks.map(roadAddressLinkToRoadAddress(false))
+      val addresses = oldAddressLinks.map(roadAddressLinkToRoadAddress(false)).map(i => i.copy(commonHistoryId = 123456))
```
leirasf: Why don't we use  the second parameter of `roadAddressLinkToRoadAddress` instead of map all the values after?

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/process/RoadAddressChangeInfoMapperSpec.scala
```diff
@@ -33,6 +33,7 @@ class RoadAddressChangeInfoMapperSpec extends FunSuite with Matchers {
     results.values.flatten.exists(_.endAddrMValue == 1000) should be (true)
     results.values.flatten.exists(_.endAddrMValue == 1400) should be (true)
     results.values.flatten.forall(_.adjustedTimestamp == 96400L) should be (true)
+    results.values.flatten.map(_.commonHistoryId).toSet.size should be (1)
```
leirasf: I think we should check if it still have the same value after the process execute. Not only check if the value is equal. Because let's imagine that we generate a new common history id for all, the test will pass but and it's not suppose.

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/process/RoadAddressChangeInfoMapperSpec.scala
```diff
@@ -64,6 +65,7 @@ class RoadAddressChangeInfoMapperSpec extends FunSuite with Matchers {
     results.values.flatten.exists(_.startAddrMValue == 1000) should be (true)
     results.values.flatten.exists(_.endAddrMValue == 1000) should be (true)
     results.values.flatten.exists(_.endAddrMValue == 1400) should be (true)
+    results.values.flatten.map(_.commonHistoryId).toSet.size should be (1)
```
leirasf: I think we should check if it still have the same value after the process execute. Not only check if the value is equal. Because let's imagine that we generate a new common history id for all, the test will pass but and it's not suppose.

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/process/RoadAddressChangeInfoMapperSpec.scala
```diff
@@ -102,6 +104,7 @@ class RoadAddressChangeInfoMapperSpec extends FunSuite with Matchers {
     results(roadLinkId3).size should be (1)
     results(roadLinkId3).count(_.id == -1000) should be (0)
     results(roadLinkId3).head.eq(roadAddress3) should be (true)
+    results.values.flatten.map(_.commonHistoryId).toSet.size should be (1)
```
leirasf: I think we should check if it still have the same value after the process execute. Not only check if the value is equal. Because let's imagine that we generate a new common history id for all, the test will pass but and it's not suppose.

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/ProjectServiceSpec.scala
```diff
@@ -1236,4 +1236,70 @@ class ProjectServiceSpec extends FunSuite with Matchers with BeforeAndAfter {
...
+    val result = projectService.createSplitRoadAddress(roadAddress, transferAndNew, project)
+    result should have size(4)
+    result.count(_.terminated == TerminationCode.Termination) should be (1)
+    result.count(_.startDate == roadAddress.startDate) should be (2)
+    result.count(_.startDate.get == project.startDate) should be (2)
+    result.count(_.endDate.isEmpty) should be (2)
+    result.filter(_.terminated == TerminationCode.NoTermination).foreach(r => r.commonHistoryId should be (-1000))
+  }
+
+  ignore("splitting road address and saving with different TRACK on transfer + new creates one new commonHistoryId for the segment") {
```
oappicgi: why are these tests ignored?

Campanha: those two ignored tests should be uncommented in next sprint where the validations from user changes will need to be checked. I will remove them since we dont need them yet.

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/process/RoadAddressChangeInfoMapperSpec.scala
Campanha: In that cases we would probably have rows with old commonHistoryIds and rows with newCommonHistoryIds that would make the set to have 2 different commonHistoryId results. But ive still made the change as it would make the code more readable.

PR #1142, opened 2018-02-08
Viite 1093 viite has 43 floating road addresses

plourenco90:


digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/RoadAddressService.scala
```diff
@@ -651,11 +650,11 @@ class RoadAddressService(roadLinkService: RoadLinkService, eventbus: DigiroadEve
       rl.linkId -> buildRoadAddressLink(rl, ra, Seq())
     }.toMap
 
-    val (filledTopology, changeSet) = RoadAddressFiller.fillTopology(allRoadLinks, viiteRoadLinks)
+    val (filledTopology, changeSet) = RoadAddressFiller.fillTopology(roadLinksWithComplementary, viiteRoadLinks)
 
     publishChangeSet(changeSet)
 
-    filledTopology
+    filledTopology ++ suravageLinks
```
leirasf: If we don't use the suravage links on the fetch, they will not be returned anymore. Probably what we want is to don't have missing road address on top of suravage links.

digiroad2-viite/src/main/scala/fi/liikennevirasto/viite/util/AssetDataImporter.scala
```diff
@@ -307,8 +307,8 @@ class AssetDataImporter {
               val newGeom = GeometryUtils.truncateGeometry3D(roadLink.geometry, segment.startMValue, segment.endMValue)
             if (!segment.geometry.equals(Nil) && !newGeom.equals(Nil)) {
 
-              if (((segment.geometry.head.distance2DTo(newGeom.head) > 1) && (segment.geometry.head.distance2DTo(newGeom.last) > 1)) ||
-                ((segment.geometry.last.distance2DTo(newGeom.head) > 1) && (segment.geometry.last.distance2DTo(newGeom.last) > 1))) {
+              if (((segment.geometry.head.distance2DTo(newGeom.head) > 0.5) && (segment.geometry.head.distance2DTo(newGeom.last) > 0.5)) ||
```
skosonen: 0.5 could be defined as constant

PR #1143, opened 2018-02-08
Droth 1166 pointlike assets dont get verified stamp & Searchbox fix

oskarmalmstrom:


UI/src/model/TrafficSignsCollection.js
```diff
@@ -65,7 +65,8 @@
         .then(function(assets) {
           eventbus.trigger('pointAssets:fetched');
           me.allowComplementaryIsActive(allowComplementary);
-            return filterTrafficSigns(me.filterComplementaries(assets));
+          verificationCollection.fetch(boundingBox, 300);
+          return filterTrafficSigns(me.filterComplementaries(assets));
```
joanacoelho: Probably we can pass to the collection the spec that has the layerName and the allowComplementary in order to have the type id. (Just to not put the 300 ). Like is being done on pointassetCollection

UI/src/application.js
```diff
@@ -23,7 +23,7 @@
     });
 
     var pointAssets = _.map(pointAssetSpecs, function(spec) {
-      var collection = _.isUndefined(spec.collection ) ?  new PointAssetsCollection(backend, spec.layerName, spec.allowComplementaryLinks) : new spec.collection(backend, spec.layerName, spec.allowComplementaryLinks) ;
+      var collection = _.isUndefined(spec.collection ) ?  new PointAssetsCollection(backend, spec.layerName, spec.allowComplementaryLinks, verificationCollection, spec.typeId) : new spec.collection(backend, spec.layerName, spec.allowComplementaryLinks, verificationCollection) ;
```
joanacoelho: Probably we can just pass through the spec, instead of separate every variable. What I mean is: 

`new PointAssetsCollection(backend, spec, verificationCollection) `

joanacoelho:
The comments are just improvments, but looks good :)

PR #1144, opened 2018-02-08
VIITE-1115_ely-code from db when possible

oappicgi:
change to reoadaddresslinkbuilder methods to take road address ely instead of making a guess


oappicgi:
added do not merge, but feel free if you think it is ok. 
I didn't check actual data from viite-api so thats why i added do not merge, but wanted to let everyone else check if this is ok

PR #1146, opened 2018-02-09
VIITE-1119_conversion-table has to be defined in roadaddressimport

oappicgi:
added requirment for conversiontable name as input to importroadaddress batch.
will be read as second argument

### DIFF
--- a/digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/util/DataFixture.scala
+++ b/digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/util/DataFixture.scala
@@ -31,9 +31,6 @@ object DataFixture {
     props
   }
 
-  lazy val konvTableNameProd = "VVH_TIEOSOITE_TAMMI2018"
-
-  lazy val konvTableNameDev = konvTableNameProd // TODO change to for e.g. "vvh_tiehistoria_heina2017_devview"
 
   val dataImporter = new AssetDataImporter
   lazy val vvhClient: VVHClient = {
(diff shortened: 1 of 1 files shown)


oappicgi:
hmm doesnt seem to work in task runner (even though it works from command line)

digiroad2-viite/src/test/scala/fi/liikennevirasto/viite/util/DataFixture.scala
```diff
@@ -77,29 +74,32 @@ object DataFixture {
...
-      val importOptions = ImportOptions(
-        onlyComplementaryLinks = false,
-        useFrozenLinkService = dr2properties.getProperty("digiroad2.VVHRoadlink.frozen", "false").toBoolean,
-        geometryAdjustedTimeStamp.toLong, conversionTable,
-        onlyCurrentRoads = dr2properties.getProperty("digiroad2.importOnlyCurrent", "false").toBoolean)
-      dataImporter.importRoadAddressData(Conversion.database(), vvhClient, importOptions)
+      importTableName match {
+        case None => // shouldn't get here because args size test
+          println(s"****** Import failed! conversiontable name required as second input ******")
+          System.exit(-1)
```
hestrompa: Can we avoid this System.exit(-1)? If we just return nothing I think the result will be the same

oappicgi: k, thought it wouldn't show as error in jenkins if it didn't throw -1. sure i can remove it

hestrompa: I think that it is not wrong at all, I just think we can handle it in a different way :) 
Like using the break() function or returning an exception.

oappicgi: ok, thanks i will add that =)

PR #1148, opened 2018-02-12
Viite 1018 track code validations

plourenco90:


oappicgi:
i might be mistaken, but i am not sure if this works with following case
      /-----------------\               /----------------               /--------------\
---/                            \--------/                          \-------/                         \_ ________
     \__________________/             \________________/            \______________-/


one can be found from  18708  part 1

plourenco90:
I did a small fix in the validation that should cover those cases.

PR #1158, opened 2018-02-16
Viite 702, 1125 and 1127 preserve the road address history

hestrompa:


hestrompa:
Tests are running ok after merging master to branch

PR #1164, opened 2018-03-07
Droth 1207 search with the passenger id for the bus stops

oskarmalmstrom:


digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -259,31 +259,32 @@ class Digiroad2Api(val roadLinkService: RoadLinkService,
...
+            "nationalId" -> stop.nationalId,
+            "stopTypes" -> stop.stopTypes,
+            "lat" -> stop.lat,
+            "lon" -> stop.lon,
+            "validityDirection" -> stop.validityDirection,
+            "bearing" -> stop.bearing,
+            "validityPeriod" -> stop.validityPeriod,
+            "floating" -> stop.floating,
+            "propertyData" -> stop.propertyData,
+            "municipalityName" -> stop.municipalityName.getOrElse(""),
```
oskarmalmstrom: We could send only the national Id, lon, lat, or whatever is needed for the search and add the others later if needed.

oskarmalmstrom:
ok

UI/src/model/LocationSearch.js
```diff
@@ -156,6 +156,27 @@
...
+    var  massTransitStopPassengerIdSearch = function(input) {
+      return $.when(backend.getMassTransitStopByPassengerIdForSearch(input.text)).then(function(result) {
+        var toCoordinates = function (r) {
+          var title = input.text + ', ' + r.municipalityName;
+          return {title: title, lon: r.lon, lat: r.lat, nationalId: r.nationalId, resultType: "Mtstop"};
+        };
+
+        if (result.length > 0)
+          return _.map(result, toCoordinates);
+        else
```
leirasf: The else is not needed

leirasf: Else can be removed

PR #1170, opened 2018-03-27
Droth 1256 problem when trying to add traffic signs

joanacoelho:


UI/src/application.js
```diff
@@ -261,7 +261,7 @@
     });
 
     _.forEach(pointAssets, function(pointAsset ) {
-     PointAssetForm.initialize(pointAsset.typeId, pointAsset.selectedPointAsset, pointAsset.collection, pointAsset.layerName, pointAsset.formLabels, pointAsset.editConstrains || function() {return false;}, roadCollection, applicationModel, backend);
+      new PointAssetForm(pointAsset.typeId, pointAsset, pointAsset.editConstrains || function() {return false;}, roadCollection, applicationModel, backend);
```
oskarmalmstrom: is there any reason why the typeId is passed separately instead of defining it in the form like the others?

joanacoelho: No 👍

PR #1172, opened 2018-03-28
Droth 1187 railroad service road to be vizualised according to the huoltovastuu field

oskarmalmstrom:


UI/src/assetTypeConfiguration.js
```diff
@@ -478,7 +478,7 @@
           {'name': "Tarkistettu", 'propType': 'checkbox', 'id': "huoltotie_tarkistettu", value: [{typeId: 0, title: 'Ei tarkistettu'}, {typeId: 1, title: 'Tarkistettu'}]}],
         style: new ServiceRoadStyle(),
         label : new ServiceRoadLabel(),
-        isVerifiable: true
+        isVerifiable: false
```
joanacoelho: It's suppose to change this, or was a mistake when testing?

UI/src/view/linear_asset/serviceRoadLayer.js
```diff
@@ -0,0 +1,29 @@
+(function(root) {
+    root.ServiceRoadLayer  = function(params) {
+        LinearAssetLayer.call(this, params);
+        var me = this;
+        var style = params.style,
+            collection = params.collection;
+
+        var isResponsibilityStyle = true;
```
joanacoelho: Can we change this name for something more clear?

UI/src/assetTypeConfiguration.js
oskarmalmstrom: no, even though the asset is verifiable it was not included in the user story for the verification button. so the button did nothing before.

UI/src/application.js
```diff
@@ -280,6 +280,31 @@
     };
 
     var linearAssetLayers = _.reduce(linearAssets, function(acc, asset) {
+      if(asset.layerName == 'maintenanceRoad') {
+        acc[asset.layerName] = new ServiceRoadLayer({
```
leirasf: The layer could be setted in the asset type configuration

PR #1185, opened 2018-04-19
Droth 1113 separate oth from viite

leirasf:


leirasf:
Will need to wait until the DROTH-1276 is deploy to master

PR #1203, opened 2018-05-21
DROTH-1340 user access fix

oskarmalmstrom:
-municipality user is now defined by: empty roles, or only premium


marquesfm:
ok for me

PR #1204, opened 2018-05-22
Droth 1321 modifications for municipality page

oskarmalmstrom:


digiroad2-oracle/src/main/resources/db/migration/V156_2__add_column_verifiable_to_asset_type.sql
```diff
@@ -3,4 +3,4 @@
 
  UPDATE ASSET_TYPE
  SET VERIFIABLE = 1
- WHERE ID IN (20, 30, 40, 50, 60, 70, 80, 90, 100, 120, 200, 280, 300);
\ No newline at end of file
+ WHERE ID IN (20, 30, 40, 50, 60, 70, 80, 90, 100, 120, 200, 280);
```
MAC-CGI: The actual version of flyway is greater than V156_2 , this will not be executed, please remove the changes on this file

joanacoelho: The change that was done in this flyway was never executed in staging, since the value 300 was added after. So is not need to remove the changes on this file

marquesfm:
Ok for me

PR #1207, opened 2018-05-25
Droth 1350 european roads exit numbers mandatory field

oskarmalmstrom:


UI/src/view/linear_asset/linearAssetForm.js
```diff
@@ -107,11 +107,11 @@
...
     var limitValueButtons = function() {
       var separateValueElement =
-        formElements.singleValueElement(selectedLinearAsset.getValue(), "a") +
-        formElements.singleValueElement(selectedLinearAsset.getValue(), "b");
+        formElements.singleValueElement(selectedLinearAsset.getValue(), "a", selectedLinearAsset.getId()) +
+        formElements.singleValueElement(selectedLinearAsset.getValue(), "b", selectedLinearAsset.getId());
       var valueElements = selectedLinearAsset.isSplitOrSeparated() ?
         separateValueElement :
-        formElements.singleValueElement(selectedLinearAsset.getValue());
+        formElements.singleValueElement(selectedLinearAsset.getValue(), undefined, selectedLinearAsset.getId());
```
joanacoelho: do you need to pass as side code undifined? You can just do it with an empty string

UI/src/view/linear_asset/piecewiseLinearAssetFormElements.js
```diff
@@ -45,9 +45,9 @@
       }
     }
 
-    function singleValueEditElement(currentValue, sideCode, input) {
-      var withoutValue = _.isUndefined(currentValue) ? 'checked' : '';
-      var withValue = _.isUndefined(currentValue) ? '' : 'checked';
+    function singleValueEditElement(currentValue, sideCode, input, id) {
+      var withoutValue = _.isNull(id) ? 'checked' : '';
```
joanacoelho: This works on creation? Because when the asset is created we don't have id

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/linearasset/TextValueLinearAssetService.scala
```diff
@@ -122,4 +122,15 @@ class TextValueLinearAssetService(roadLinkServiceImpl: RoadLinkService, eventBus
       }
     }
   }
+
+  override def validateAssetValue(values: Option[Value]): Unit = {
+    val euroPattern = "^[0-9|Ee][0-9|Bb]{0,2}"
```
joanacoelho: Since this function is on TextualService this name could be change, to not be just about the european ones

UI/src/view/linear_asset/piecewiseLinearAssetFormElements.js
oskarmalmstrom: this is called for form creation after selecting an asset, not on value input on the form. the thing is that there are assets without values, so the check needs to be on id

UI/src/view/linear_asset/linearAssetForm.js
oskarmalmstrom: yes, the check is if (_.isUndefined(sideCode)) {}else..

joanacoelho: It could be changed to _isEmpty

marquesfm:
Ok for me

PR #1213, opened 2018-06-04
Droth 51 oth application feedback functionality

marquesfm:


digiroad2-oracle/src/main/resources/db/migration/V167__add_table_to_feedback.sql
```diff
@@ -0,0 +1,12 @@
...
+  id number primary key,
+  receiver varchar2(128),
+  createdBy varchar2(128),
+  createdAt timestamp default current_timestamp not null,
+  subject varchar2(128),
+  body varchar2(4000),
+  status char(1) default '0',
+  statusDate timestamp default current_timestamp,
+  CONSTRAINT status check (status in ('1','0'))
+);
```
MAC-CGI: could be replace by

create table feedback
(
  id number primary key,
  receiver varchar2(128),
  created_By varchar2(128),
  created_Date timestamp default current_timestamp not null,
  subject varchar2(128),
  body varchar2(4000),
  status char(1) default '0',
  status_Date timestamp default current_timestamp,
  CONSTRAINT status check (status in ('1','0'))
);

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/util/EmailAuthPropertyReader.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  lazy val properties: Properties = {
+    val props = new Properties()
+    props.load(getClass.getResourceAsStream("/keys.properties"))
+    props
+  }
+
+  def getUsername: String = {
+    val loadedKeyString = properties.getProperty("email.username")
+    println("u = "+loadedKeyString)
```
MAC-CGI: maybe this println could be removed

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/feedback/FeedbackDao.scala
```diff
@@ -0,0 +1,75 @@
...
+      FeedbackInfo(id, receiver, createdBy, createdAt, body, subject, status, statusDate)
+    }
+  }
+
+  def getAllFeedbacks(): Seq[FeedbackInfo] = {
+    val query =  s"""
+            select id, receiver, created_by, created_date, subject, body, status, status_date
+            from feedback
+            """
+      StaticQuery.queryNA[FeedbackInfo](query).iterator.toSeq
```
marquesfm: Since we are using the same select query 3 times why we don't create a getAllFeedbacks with an optional "where" input parameter.
So, each time we want using we call the getAllFeedbacks with different "where" condition.

PR #1221, opened 2018-06-26
Droth 1113 separate oth from viite 

leirasf:


leirasf:
The road address DAO, is comment on purpose.

PR #1230, opened 2018-07-09
Droth 1434/1435/1436  serviceroad style form mass update fix

oskarmalmstrom:


UI/src/view/linear_asset/serviceRoadStyle.js
```diff
@@ -21,6 +21,21 @@
...
+        return _.merge({}, linearAsset, { expired: expired }, type);
+      });
+      var offsetBySideCode = function(linearAsset) {
+        return GeometryUtils.offsetBySideCode(applicationModel.zoom.level, linearAsset);
+      };
+      var linearAssetsWithAdjustments = _.map(linearAssetsWithType, offsetBySideCode);
+      return _.sortBy(linearAssetsWithAdjustments, function(asset) {
+        return asset.expired ? -1 : 1;
+      });
+    };
```
MAC-CGI: Is there any reason for override this method getNewFeatureProperties instead of replace 'experied' to 'hasvalue' in ServiceRoadStyle?

UI/src/view/dynamicAssetForm.js
```diff
@@ -655,6 +655,13 @@
       var rootElement = $('#feature-attributes');
       _assetTypeConfiguration = assetTypeConfiguration;
 
+      var updateStatusForMassButton = function(element) {
+        if(assetTypeConfiguration.selectedLinearAsset.isSplitOrSeparated()) {
+          element.prop('disabled', !(me.isSaveable() && me.isSplitOrSeparatedAllowed()));
+        } else
+          element.prop('disabled', !(me.isSaveable()));
+      };
+
```
MAC-CGI: In **else** condition don't we need to user the dirty information ? like it is done on 
`            var updateStatus = function(element) {
                if(assetTypeConfiguration.selectedLinearAsset.isSplitOrSeparated()) {
                    element.prop('disabled', !(me.isSaveable() && me.isSplitOrSeparatedAllowed()));
                } else
                    element.prop('disabled', !(me.isSaveable() && assetTypeConfiguration.selectedLinearAsset.isDirty()));
            };` 

**Note** if it was needed we can reuse the same method

oskarmalmstrom: Tallenna should be enabled when both On/Off is checked. If switching to On and then Off, it won't be enabled with that condition.

UI/src/view/linear_asset/serviceRoadStyle.js
oskarmalmstrom: Using the one on the base class and changing the rules to 'hasAsset' onServiceRoadStyle also changes the way the overlay lines are drawn. From my experience it is not worth to start changing everything, and better to just override it here.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/linearasset/TextValueLinearAssetService.scala
```diff
@@ -127,8 +127,8 @@ class TextValueLinearAssetService(roadLinkServiceImpl: RoadLinkService, eventBus
     val euroAndExitPattern = "^[0-9|Ee][0-9|Bb]{0,2}"
     values.get match {
       case TextualValue(textValue) => textValue.split(",").forall(_.trim.matches(euroAndExitPattern)) match {
-          case true => print(textValue)
           case false => throw new AssetValueException(textValue)
+          case _ =>
```
joanacoelho: shouldn't this been taken care?

oskarmalmstrom: The previous truthful case is bad, just printing for test purposes. This a correction to pass the validation

marquesfm: At least send None, to avoid warnings.

joanacoelho: You can use `if` and `else` if it helps, something like: 

`values.get match {
      case TextualValue(textValue) =>
        if(!textValue.split(",").forall(_.trim.matches(euroAndExitPattern)))
          throw new AssetValueException("incorrect asset type")
      case _ => throw new AssetValueException("incorrect asset type")`

PR #1234, opened 2018-07-16
Droth 1413 source code for assests

oskarmalmstrom:


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -1214,14 +1214,15 @@ object DataFixture {
...
-        username.exists(user => user.toLowerCase.startsWith("a")) ||
-        username.exists(user => user.toLowerCase.startsWith("u"))
-    }
+  private def isKIdentifier(username: Option[String]): Boolean = {
+    username.exists(user => user.toLowerCase.startsWith("k")) ||
+      username.exists(user => user.toLowerCase.startsWith("lx")) ||
+      username.exists(user => user.toLowerCase.startsWith("a")) ||
+      username.exists(user => user.toLowerCase.startsWith("u"))
+  }
+
```
oskarmalmstrom: maybe we could use something like: 
    def isKIdentifier(username: Option[String]): Boolean = {
      val identifiers: Set[String] = Set("k", "lx", "a", "u")
      username.exists(user => identifiers.exists(identifier => user.toLowerCase.startsWith(identifier)))
    }

PR #1242, opened 2018-07-24
Droth 1449 turning restrictions enable fixing assets that are against traffic direction

joanacoelho:


digiroad2-oracle/src/main/resources/db/migration/V175__add_traffic_sign_to_manoeuvre.sql
```diff
@@ -0,0 +1,2 @@
+alter table manoeuvre
+add traffic_sign_id number constraint fk_traffic_sign_id references asset(id);
```
joanacoelho: This file can be deleted from this branch

PR #1246, opened 2018-07-27
Droth 1468 last updated by vvh generated timestamp should not be considered as checked for assets

joanacoelho:


joanacoelho:
ok

PR #1248, opened 2018-08-08
DROTH-1466 related fix warning

joanacoelho:


joanacoelho:
Resolves warnings that appears when compiling the project: 

Warning reflective access of structural type member method should be enabled by making the implicit value scala.language.reflectiveCalls visible.

PR #1250, opened 2018-08-14
Droth 1437 optional save for bus stops

oskarmalmstrom:


UI/src/view/point_asset/massTransitStopForm.js
```diff
@@ -67,14 +75,31 @@
...
+            successCallback: function () {
+              selectedMassTransitStopModel.setAdditionalProperty('trSave', [{ propertyValue: 'false' }]);
+              saveStop();
+            },
+            closeCallback: function () {
+              saveStop();
+            }
+          });
+        } else {
+          saveStop();
```
joanacoelho: The `saveStop()` call can not be done on the end of everything instead of being inside of `successCallback`, `closeCallback` and the `else ` ?

oskarmalmstrom: Maybe in some way, but now if its outside of all brackets it calls the savestop() before answering.

PR #1271, opened 2018-09-21
DROTH-1571 When logging as a municipality user the focus doesn't work

joanacoelho:


digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -189,14 +189,13 @@ class Digiroad2Api(val roadLinkService: RoadLinkService,
...
       case _ =>
-        if(user.isMunicipalityMaintainer())
+        if(user.isMunicipalityMaintainer() && user.configuration.authorizedMunicipalities.nonEmpty )
           getStartUpParameters(user.configuration.authorizedMunicipalities, municipalityDao.getCenterViewMunicipality)
         else {
-          if (user.isServiceRoadMaintainer())
+          if (user.isServiceRoadMaintainer() && user.configuration.authorizedAreas.nonEmpty)
             getStartUpParameters(user.configuration.authorizedAreas, municipalityDao.getCenterViewArea)
-          else if (user.isBusStopMaintainer()) //case ely maintainer
+          else if (user.isBusStopMaintainer() && user.configuration.authorizedAreas.nonEmpty) //case ely maintainer
```
marquesfm: Should't be "user.configuration.authorizedMunicipalities.nonEmpty" since inside the if we are using user.configuration.authorizedMunicipalities?

PR #1284, opened 2018-10-09
Droth 1556 changing zoom level and labels

joanacoelho:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/MassTransitStopDao.scala
```diff
@@ -63,6 +63,23 @@ class MassTransitStopDao {
     queryToPersistedMassTransitStops(queryFilter(query))
   }
 
+  def fetchLightGeometry(queryFilter: String => String): Seq[LightGeometryMassTransitStop] = {
+    val query =
+      """
+        select a.geometry, a.valid_from, a.valid_to
+        from asset a
+          join asset_link al on a.id = al.asset_id
+          join lrm_position lrm on al.position_id = lrm.id
```
MAC-CGI: lrm_position table should be removed

PR #1286, opened 2018-10-11
Droth 1595 tierekisteri import improvement

oskarmalmstrom:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/client/tierekisteri/importer/TrafficSignTierekisteriImporter.scala
```diff
@@ -68,7 +68,6 @@ class TrafficSignTierekisteriImporter extends PointAssetTierekisteriImporterOper
         point =>
           val trafficSign = IncomingTrafficSign(point.x, point.y, vvhRoadlink.linkId, generateProperties(trAssetData),
             getSideCode(roadAddress, trAssetData.track, trAssetData.roadSide).value, Some(GeometryUtils.calculateBearing(vvhRoadlink.geometry)))
-          println(s"Created OTH $assetName asset on link ${vvhRoadlink.linkId} from TR data")
```
marquesfm: Don't' remove that information, its useful, but moved to after the create method outside of the map.

PR #1335, opened 2019-01-15
DROTH-1783_experimental_zoom_lvl_fix

oskarmalmstrom:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/linearasset/OracleLinearAssetDao.scala
```diff
@@ -444,7 +446,7 @@ class OracleLinearAssetDao(val vvhClient: VVHClient, val roadLinkService: RoadLi
            join property p on p.public_id = 'mittarajoitus'
            left join administrative_class ad on pos.link_id = ad.link_id
            left join number_property_value s on s.asset_id = a.id and s.property_id = p.id
-           where a.floating = 0 and ad.valid_to is null and a.asset_type_id = #$assetTypeId #$linkGeomCondition"""
+           where a.floating = 0 and ad.valid_to is null and a.asset_type_id = #$assetTypeId #$linkGeomCondition and #$boundingBoxFilter"""
```
oskarmalmstrom: This fetch is taking some time. Do you think it would make sense to/be faster to exclude certain fields on (a.verified_by, a.verified_date, s.value) from the query?

PR #1353, opened 2019-02-18
speed limit test modifications

oskarmalmstrom:


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/dao/linearasset/OracleSpeedLimitDaoSpec.scala
```diff
@@ -81,7 +82,7 @@ class OracleSpeedLimitDaoSpec extends FunSuite with Matchers {
 
       val speedLimits = dao.getSpeedLimitLinksByRoadLinks(roadLinks)
 
-      speedLimits._1.map(_.id) should equal(Seq(300103))
+      speedLimits._1.size should equal(1)
```
MAC-CGI: Isn't enough validate only the size, if the size is 1 you should validate the id using head

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/dao/linearasset/OracleSpeedLimitDaoSpec.scala
```diff
@@ -96,7 +97,7 @@ class OracleSpeedLimitDaoSpec extends FunSuite with Matchers {
...
```
MAC-CGI: Isn't enough validate only the size, if the size is 1 you should validate the id using head

marquesfm:
The problem of these tests can be easily fixed if we modify the municipality number on "Get speed_limits requires municipality number" test to one not used on other unit tests, like for example 588.

oskarmalmstrom:
integration api spec changed to perform get on different municipality to not interfere(fill topology etc) with test data

PR #1359, opened 2019-02-22
DROTH-1707 Batch now executable by sign group

oskarmalmstrom:


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -1507,7 +1510,7 @@ object DataFixture {
         }
       }
       additionalPanelIdToExpire.foreach { case (id, linkId, signType) =>
-        trafficSignService.expireAssetWithoutTransaction(trafficSignService.withIds(Set(id).flatten), Some("batch_process_panel_merge"))
+//        trafficSignService.expireAssetWithoutTransaction(trafficSignService.withIds(Set(id).flatten), Some("batch_process_panel_merge"))
```
moraispedro: Since it's commented, can it be removed?

oskarmalmstrom: should be commented, we will now run the merges for all the signs and as the last measure expire all the panels

PR #1374, opened 2019-03-22
component_removal_1648

oskarmalmstrom:


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -1936,8 +1898,6 @@ object DataFixture {
         }
       case Some("create_traffic_signs_using_linear_assets") =>
```
marquesfm: Confirm this part, since we don't do creations of traffic Signs using linear, we only do the opposite for now.
And I don't know if the createProhibitionsUsingTrafficSigns() is not used on the future batch.

PR #1380, opened 2019-03-29
Droth-1838_search_tiekunta_bug_fixes

moraispedro:


UI/src/view/navigationpanel/searchBox.js
```diff
@@ -47,29 +47,10 @@
 
       var showDialog = function(message) {
         instructionsPopup.show(message, 3000);
+        jQuery('#search-results').parent().hide();
```
oskarmalmstrom: what happens with the existing spinner?

UI/src/view/navigationpanel/searchBox.js
```diff
@@ -47,29 +47,11 @@
 
       var showDialog = function(message) {
         instructionsPopup.show(message, 3000);
+        jQuery('.spinner-overlay-search').remove();
+        jQuery('#search-results').parent().hide();
       };
       locationSearch.search(coordinatesText.val()).then(function(results) {
         populateSearchResults(results);
```
MAC-CGI: Please, check if this code is not even necessary. I think when we only have one result, the map should automatically move

PR #1387, opened 2019-04-09
Droth 1860 publishing xml for bus stop change info

moraispedro:


gruntfile.js
```diff
@@ -62,73 +62,73 @@ module.exports = function(grunt) {
     },
```
oskarmalmstrom: should probably be reverted?

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/ChangeApi.scala
```diff
@@ -315,6 +341,86 @@ class ChangeApi(val swagger: Swagger) extends ScalatraServlet with JacksonJsonSu
...
+              municipalityService.getMunicipalitiesNameAndIdByCode(Set(stop.pointAsset.municipalityCode))
+            }
+          val massTransitStop = pointAsset.asInstanceOf[PersistedMassTransitStop]
+          val busStopTypes = getPropertyValuesByPublicId("pysakin_tyyppi", massTransitStop.propertyData).map(x => x.propertyValue.toLong).toSet
+          val modificationInfo = massTransitStop.modified.modificationTime match {
+            case Some(_) => massTransitStop.modified
+            case _ => massTransitStop.created
+          }
+          val validTo = getValiditiDatesProperties(massTransitStop, "viimeinen_voimassaolopaiva")
+          val validFrom = getValiditiDatesProperties(massTransitStop, "ensimmainen_voimassaolopaiva")
```
moraispedro: Validity instead of validiti.

gruntfile.js
moraispedro: I think this was my mistake, when commiting some minor changes, accidentally included the gruntfile, that was then reverted. Just check if it's all okay.

PR #1407, opened 2019-06-17
DROTH-2021 Wrong id on the feedback form for bus stops

moraispedro:


UI/src/controller/feedbackModel.js
```diff
@@ -47,7 +47,8 @@
                 }
                 else {
                     linkId = selected.linkId ? [selected.linkId] : _.map(selected.assets, function (asset) { return asset.linkId; }).filter(Boolean);
-                    assetId = selected.id ? [selected.id] : _.map(selected.assets, function (asset) { return asset.id; }).filter(Boolean);
+                    assetId = assetInfo.typeId === me.assetConfig.assetTypes.massTransitStop ?
```
oskarmalmstrom: The use of a ternary operator when we are just dealing with two alternatives(mass transit stop/other), but becomes a bit messy if we need to add more. Maybe a switch or if else would be clearer?

PR #1408, opened 2019-06-19
DROTH-2027 (US 1863) It's possible set starting date later than end date when edit

marquesfm:
- Added validation to avoid starting date later than the end date on roadDamagedByThaw and roadWorks assets.

### DIFF
--- a/UI/src/assetTypeConfiguration.js
+++ b/UI/src/assetTypeConfiguration.js
@@ -50,6 +50,22 @@
       trWeightGroup: [assetType.trWeightLimits, assetType.trTrailerTruckWeightLimits, assetType.trAxleWeightLimits, assetType.trBogieWeightLimits]
     };
 
+    var datePeriodValueExtract = function (date) {
+      var datePeriodValue = date.getPropertyValue().values;
+      var startDate = new Date(_.head(datePeriodValue).value.startDate.replace(/(\d+).(\d+).(\d{4})/, "$2/$1/$3"));
+      var endDate = new Date(_.head(datePeriodValue).value.endDate.replace(/(\d+).(\d+).(\d{4})/, "$2/$1/$3"));
+
+      return [startDate, endDate];
+    };
+
+    var isEndDateAfterStartdate = function (date) {
+      var datePeriods = datePeriodValueExtract(date);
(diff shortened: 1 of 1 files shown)


UI/src/assetTypeConfiguration.js
```diff
@@ -50,6 +50,22 @@
...
 
+    var datePeriodValueExtract = function (date) {
+      var datePeriodValue = date.getPropertyValue().values;
+      var startDate = new Date(_.head(datePeriodValue).value.startDate.replace(/(\d+).(\d+).(\d{4})/, "$2/$1/$3"));
+      var endDate = new Date(_.head(datePeriodValue).value.endDate.replace(/(\d+).(\d+).(\d{4})/, "$2/$1/$3"));
+
+      return [startDate, endDate];
+    };
+
+    var isEndDateAfterStartdate = function (date) {
```
moraispedro: I believe this can be optimized to:

```suggestion
    var isEndDateAfterStartdate = function (date) {
		_.reduce(datePeriodValueExtract(date), function(startDate, endDate){ return startDate <= endDate;})
	};
```

UI/src/assetTypeConfiguration.js
```diff
@@ -285,19 +301,23 @@
           var datePeriodField = _.filter(fields, function(field) { return field.getPropertyValue().propertyType === 'date_period'; });
 
           var isInDatePeriod = function(date) {
```
moraispedro: Same as before:
```suggestion
			var isInDatePeriod = function(date) {
				_.reduce(datePeriodValueExtract(date), function(startDate, endDate){ 
					return new Date(endDate.getMonth() + '/' + endDate.getDate() + '/' + (endDate.getFullYear() - 1)) <= startDate;
				});
			};
```

Check first.

UI/src/assetTypeConfiguration.js
marquesfm: I choose Miguel solution! 👍

UI/src/assetTypeConfiguration.js
marquesfm: I choose Miguel solution! 👍

PR #1413, opened 2019-07-02
Droth 1987 adding assets to tnits

marquesfm:


gruntfile.js
```diff
@@ -62,73 +62,73 @@ module.exports = function(grunt) {
     },
     clean: ['dist'],
     connect: {
```
moraispedro: Gruntfile was changed. Confirm if it's correct, since it was commited by mistake.

PR #1421, opened 2019-08-23
DROTH-2092 Create Batch to get Wrong functional classes on the roads

marquesfm:
- Added new batch name on all environments pipelines;
- New method added at RoadLinkDAO to get all RoadLinks with overridden specific functional class value since a specific date;
- Added new batch to get all RoadLinks with overridden specific functional class value since a specific date.


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -1770,6 +1770,52 @@ object DataFixture {
...
+  def getStateRoadWithFunctionalClassOverridden(): Unit = {
+    println("\nStart process to get StateRoads With Functional Class Overridden")
+    println(DateTime.now())
+    println("")
+
+    val functionalClassValue = 5
+    val sinceDate = Some("20190101") //Format required YYYYMMDD
+
+    OracleDatabase.withDynTransaction {
+      val linkIdsOverridden = RoadLinkDAO.getLinkIdByValue("functional_class", functionalClassValue, sinceDate).toSet
```
MAC-CGI: Since FunctionalClassDao extends from RoadLinkDAO, this should be like 
 FunctionalClassDao .getLinkIdByValue(functionalClassValue, sinceDate).toSet

PR #1423, opened 2019-08-26
Droth 2013 adding liikvast info

marquesfm:


digiroad2-oracle/src/main/resources/db/migration/V210__add_new_road_sign_side_column.sql
```diff
@@ -0,0 +1 @@
+alter table asset add sign_road_side char(1) default '0';
```
oskarmalmstrom: Does it make sense to have this on the asset table instead of adding a new property?

UI/src/view/point_asset/trafficSignForm.js
```diff
@@ -43,6 +43,7 @@
       var additionalPanels = getProperties(asset.propertyData, "additional_panel");
       var checked = _.isEmpty(additionalPanels.values) ? '' : 'checked';
       var renderedPanels = checked ? renderAdditionalPanels(additionalPanels, collection) : '';
+      var roadSideInfo = asset.roadSideInfo ? "Kyllä" : "Ei";
```
oskarmalmstrom: The variable name is not descriptive. Could be 'onOppositeSide' or something similar.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/client/tierekisteri/TierekisteriTrafficSignAssetClient.scala
```diff
@@ -32,16 +32,17 @@ class TierekisteriTrafficSignAssetClient(trEndPoint: String, trEnable: Boolean,
     val roadPartNumber = convertToLong(getMandatoryFieldValue(data, trRoadPartNumber)).get
     val startMValue = convertToLong(getMandatoryFieldValue(data, trStartMValue)).get
     val track = convertToInt(getMandatoryFieldValue(data, trTrackCode)).map(Track.apply).getOrElse(Track.Unknown)
+    val wrongSideInformation = getFieldValue(data, trLIIKVAST)
```
oskarmalmstrom: Maybe not so serious but i don't like 'wrong' side. Opposite side sounds better :)

UI/src/view/point_asset/trafficSignForm.js
marquesfm: trafficFlowContraryDirection

UI/src/view/point_asset/trafficSignForm.js
```diff
@@ -52,15 +53,21 @@
       '        <label class="traffic-panel-checkbox-label">Linkitä lisäkilpiä</label>' +
       '    </div>';
 
+      var wrongSideInfo = asset.id !== 0 ?
```
marquesfm: We can switch this name too, for example to "ContraryDirectionInfo"

digiroad2-oracle/src/main/resources/db/migration/V210__add_new_road_sign_side_column.sql
marquesfm: Make more sense in the properties, because it's not related to all assets, only for Traffic Signs

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/client/tierekisteri/TierekisteriTrafficSignAssetClient.scala
marquesfm: Let's switch this name too!

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/pointasset/OracleTrafficSignDao.scala
```diff
@@ -27,7 +27,8 @@ case class PersistedTrafficSign(id: Long, linkId: Long,
                                 validityDirection: Int,
                                 bearing: Option[Int],
                                 linkSource: LinkGeomSource,
-                                expired: Boolean = false) extends PersistedPoint
+                                expired: Boolean = false,
+                                roadSideInfo: Boolean = false) extends PersistedPoint
```
marquesfm: this name can be modified too
`invertedDirection: Boolean = false) extends PersistedPoint`

UI/src/view/point_asset/trafficSignForm.js
FilPinto: oppositeSide or similar as Oskar suggested seems a better option

PR #1424, opened 2019-09-11
Droth 1893 improvements municipality page part 2

marquesfm:


UI/src/view/municipalityWorkList.js
```diff
@@ -89,13 +89,73 @@
...
+        var canvas_image_width = HTML_Width;
+        var canvas_image_height = HTML_Height;
+        var totalPDFPages = Math.ceil(HTML_Height/PDF_Height)-1;
+
+        html2canvas(document.querySelector('#pdf')).then( function (canvas){
+              var imgData = canvas.toDataURL('image/jpeg');
+          var doc = new jsPDF('p', 'pt', [PDF_Width, PDF_Height]);
+          doc.addImage(imgData, 'JPEG', top_left_margin, top_left_margin,canvas_image_width,canvas_image_height);
+
+          for (var i = 1; i <= totalPDFPages; i++) {
```
oskarmalmstrom: could be done with a _.forEach instead of traditional for loop. no difference in functionality but looks a bit better

FilPinto: I don't have any collection in the code you are commenting on so I can't use the _.foreach method.

oskarmalmstrom: true, i need to look better

UI/src/view/municipalityWorkList.js
```diff
@@ -89,13 +89,73 @@
...
+            doc.addPage(PDF_Width, PDF_Height);
```
MAC-CGI: Since this line is not change, could be move for outside of the loop

UI/src/view/municipalityWorkList.js
```diff
@@ -89,13 +89,73 @@
...
+            doc.addImage(imgData, 'JPEG', top_left_margin, -(PDF_Height*i)+(top_left_margin*4),canvas_image_width,canvas_image_height);
+          }
+
+          var date = new Date();
+            doc.save('DR' + '_' + municipalityName.toLowerCase() + '_' + String(date.getDate()).padStart(2, '0') + String(date.getMonth() + 1).padStart(2, '0') + String(date.getFullYear()) + '.pdf');
+            });
+        pdf.html('').css('height', '0px').css('width', '0px');
```
MAC-CGI: please, this css should be move to a specific 'less' file

UI/src/view/municipalityWorkList.js
FilPinto: Even though the loop doesn't change it directly by the counter it is still necessary to be there since the pages are variable (depends on the quantity of assets) and so the totalPDFPages dictate them

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/RoadLinkService.scala
```diff
@@ -791,50 +819,65 @@ class RoadLinkService(val vvhClient: VVHClient, val eventbus: DigiroadEventBus,
     }
   }
 
-    private def fetchOverridedRoadLinkAttributes(idTableName: String): List[(Long, Option[(String, String)])] = {
+  private def fetchOverridedRoadLinkAttributes(idTableName: String): List[(Long, Option[(String, String)])] = {
     val fetchResult =
-      sql"""select i.id, rla.link_id, rla.name, rla.value, rla.created_date, rla.created_by, rla.modified_date, rla.modified_by
+      sql"""select /*+ PARALLEL */ i.id, rla.link_id, rla.name, rla.value, rla.created_date, rla.created_by, rla.modified_date, rla.modified_by
```
ricardofmsimoes: Remove /*+ PARALLEL */ if it is not needed

PR #1427, opened 2019-09-19
DROTH-2099 Unknown speed limit list disappears some times

FilPinto:
-Fixed form loading first than verification

### DIFF
--- a/UI/src/view/linear_asset/speedLimitForm.js
+++ b/UI/src/view/linear_asset/speedLimitForm.js
@@ -130,9 +130,9 @@
       }
   };
 
-  var bindEvents = function(selectedSpeedLimit, feedbackCollection) {
+  var bindEvents = function(selectedSpeedLimit, feedbackCollection, mapBackend) {
     new FeedbackDataTool(feedbackCollection, 'speedLimit', authorizationPolicy);
-    backend = feedbackCollection.backend;
+    backend = mapBackend;
     if (authorizationPolicy.userRoles.length === 0){
       backend.getUserRoles();
     }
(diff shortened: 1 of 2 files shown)


UI/src/view/linear_asset/speedLimitForm.js
```diff
@@ -129,9 +129,13 @@
       }
   };
 
-  var bindEvents = function(selectedSpeedLimit, feedbackCollection) {
+  var bindEvents = function(selectedSpeedLimit, feedbackCollection, backend) {
     new FeedbackDataTool(feedbackCollection, 'speedLimit', authorizationPolicy);
 
+    if (authorizationPolicy.userRoles.length === 0){
+      backend.getUserRoles();
```
oskarmalmstrom: Another option could be to only render the links when we know that the roles have been fetched, maybe like this:
`    eventbus.on('layer:selected', function(layer) {
      if(layer === 'speedLimit') {
        $('ul[class=information-content]').empty();
        eventbus.on('roles:fetched', function() {
            renderLinktoWorkList();
        });
      }
    });`

on line 179. I tested it locally and seems to work, but not sure how it is on the real environments. Should be ok i think

FilPinto: This bug happens when reloading the page when the speedlimit asset is selected. The bug itself happens because the speed limits form is loading first than the event that triggers the 'roles:fetched' (it's the last thing the application.js does) that the authorizationPolicy uses. That option you suggested doesn't work because you will create the two kind of forms (the one for the operator and the other for the remaining of the users).

PR #1429, opened 2019-09-23
Droth 2014 modifying oth initial view

marquesfm:


UI/src/application.js
```diff
@@ -137,14 +140,35 @@
...
+        assetTypeId = startupParameters.assetType;
+        defaultLocation = [startupParameters.lon, startupParameters.lat, startupParameters.zoom];
+
+        assetTypeLayerName = getSelectedAssetByTypeId(pointAssets, assetTypeId);
+        if(typeof assetTypeLayerName === 'undefined'){
+          assetTypeLayerName = getSelectedAssetByTypeId(linearAssets, assetTypeId);
+        }
+        if(typeof assetTypeLayerName === 'undefined'){
+          assetTypeLayerName = getSelectedAssetByTypeId(groupedPointAssets, assetTypeId);
+        }
```
oskarmalmstrom: can be shortened to        ` assetTypeLayerName = getSelectedAssetByTypeId(pointAssets, assetTypeId) || getSelectedAssetByTypeId(linearAssets, assetTypeId);
`
Grouped point assets can't be selected like that. Pointlike weight limits (TR painorajoitukset) is a layer with 4 different asset types so there needs to be a different solution if you want that as the default layer.

UI/src/view/defaultLocationButton.js
```diff
@@ -1,23 +1,62 @@
...
+  root.DefaultLocationButton = function (map, container, backend, assetTypeConfig, defaultAssetType, defaultCoordinates) {
+    var vkm = function() {
+      if(zoom <= 5){
+        backend.getMunicipalityFromCoordinates(lon, lat, function (vkmResult) {
+              var municipalityInfo = vkmResult.kunta ? vkmResult.kunta : "Suomi";
+              municipality = municipalityInfo;
+              ely = vkmResult.ely_nimi ? vkmResult.ely_nimi : "Suomi";
+            }, function () {
+              municipality = 'Suomi';
+              ely = 'Suomi';
```
oskarmalmstrom: Aren't these values already assigned on line 19 and 20?

UI/src/view/defaultLocationButton.js
```diff
@@ -1,23 +1,62 @@
...
       '<div class="default-location-btn-container"></div>';
 
     container.append(element);
 
 
     eventbus.on('roles:fetched', function(userInfo) {
-      if (!(_.some(userInfo.roles, function(role) {return role === "viewer"; })))
-        $('.default-location-btn-container').append('<button class="btn btn-sm btn-tertiary" id="default-location-btn">Muuta oletussijainniksi</button>');
+      if (!(_.some(userInfo.roles, function(role) {return (role === "viewer" || role === "serviceRoadMaintainer");}))) {
+          roles = userInfo.roles[0];
```
oskarmalmstrom: What if the user has multiple roles? How can you be sure to always pick the first one?

UI/src/view/changeInitialViewPopup.js
```diff
@@ -0,0 +1,221 @@
...
+      else {
+        if (userRole !== "busStopMaintainer") {
+          defaultParameters.municipalityId = parseInt(selectedLocation, 10);
+        }
+        else {
+          defaultParameters.elyId = parseInt(selectedLocation,10);
+        }
+        backend.updateUserConfigurationDefaultLocation(defaultParameters);
+      }
+
```
oskarmalmstrom: can we call `backend.updateUserConfigurationDefaultLocation(defaultParameters);` here instead? It is called regardless of condition

UI/src/view/changeInitialViewPopup.js
```diff
@@ -0,0 +1,221 @@
...
+      setTimeout(function () {
+        new GenericConfirmPopup("Laskeutumissivu päivitetty.", {type: 'alert'});
+      }, 1);
```
oskarmalmstrom: This sets the timeout only at 1 millisecond

UI/src/view/defaultLocationButton.js
```diff
@@ -1,23 +1,62 @@
...
+          if (roles === "busStopMaintainer") {
+              places = backend.getUserElyConfiguration();
+              municipality = ely;
+          } else if (roles === "operator") {
+              places = [];
+          } else {
+              places = backend.getUnverifiedMunicipalities();
```
oskarmalmstrom: unrelated to this user story: function could be renamed to getMunicpalities or something. It only fetches the municipality names and ids, nothing to do with unverified.

UI/src/view/defaultLocationButton.js
```diff
@@ -1,23 +1,62 @@
...
-        backend.updateUserConfigurationDefaultLocation(actualLocationInfo, function () {
-          new GenericConfirmPopup("Oletussijainti tallennettu.", {type: 'alert'});
-        }, function () {
-          alert('Tarkistus epäonnistui. Yritä hetken kuluttua uudestaan.');
+
+          if (places.length === 0) {
+              new ChangeInitialViewPopup(backend, actualLocationInfo, roles, places, assetTypeConfig, defaultAsset, municipality);
+      }
+      else{
+              places.then(function (places) {
```
oskarmalmstrom: what does the ` .then` do in this case? For me it is only used with a promise on a request but i might be wrong

UI/src/view/defaultLocationButton.js
FilPinto: The users who can have a combination of roles are ely maintainers and service maintainers between themselves and the way a user is created with that combination will be ely first in roles[]

UI/src/view/changeInitialViewPopup.js
FilPinto: For some reason the popup will not stay in place without the setTimeout. If it is 1 millisecond it will stay. But i will spend a little time on this to see if i can work this in other way.

UI/src/view/defaultLocationButton.js
FilPinto: I agree. I will change that.

UI/src/view/changeInitialViewPopup.js
```diff
@@ -0,0 +1,215 @@
+window.ChangeInitialViewPopup = function(backend, location, userRole, places, assetTypeConfig, defaultAsset, municipality) {
+
+  var options = {
+    confirmButton: 'Tallenna',
+    cancelButton: 'Peruuta',
+    successCallback: function () {
+      var selectedLocation = $('#location').val();
+      var selectedAssetType = parseInt($('#assetType').val(), 10);
```
MAC-CGI: please, create a constant and replace all place where you are expecting  massTransitStopId

UI/src/view/defaultLocationButton.js
```diff
@@ -1,23 +1,58 @@
...
+          if (roles === "busStopMaintainer") {
+              places = backend.getUserElyConfiguration();
+              municipality = ely;
+          } else if (roles === "operator") {
+              places = [];
+          } else {
+              places = backend.getMunicipalities();
+          }
+          vkm();
+          $('.default-location-btn-container').append('<button class="btn btn-sm btn-tertiary" id="default-location-btn">Muokkaa aloitussivua</button>');
```
MAC-CGI: Another suggestion for this code 

switch(roles) {
  case "busStopMaintainer":
     places = backend.getUserElyConfiguration();
      municipality = ely;
    break;
  case "operator":
    places = [];
    break;
  default:
   places = backend.getMunicipalities();
}

UI/src/view/genericConfirmPopup.js
```diff
@@ -15,27 +15,27 @@ window.GenericConfirmPopup = function(message, options) {
...
-            '<div class="modal-dialog">' +
-                '<div class="content">' +
-                  message +   
-                '</div>' +
-                '<div class="actions">' +
-                    '<button class = "btn btn-primary yes">' + options.yesButtonLbl + '</button>' +
-                    '<button class = "btn btn-secondary no">' + options.noButtonLbl + '</button>' +
-                '</div>' +
-            '</div>' +
+        '<div class="modal-dialog">' +
```
MAC-CGI: please ident the code

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -212,34 +212,39 @@ class Digiroad2Api(val roadLinkService: RoadLinkService,
...
     val user = userProvider.getCurrentUser()
-    val userPreferences: Option[(Long, Long, Int)] = (user.configuration.east, user.configuration.north, user.configuration.zoom) match {
-      case (Some(east), Some(north), Some(zoom)) => Some(east, north, zoom)
+    val assetTypeId = user.configuration.assetType match {
+      case Some(assetType) => assetType
+      case _  => 10
+    }
+    val userPreferences: Option[(Long, Long, Int, Int)] = (user.configuration.east, user.configuration.north, user.configuration.zoom, user.configuration.assetType) match {
+      case (Some(east), Some(north), Some(zoom), None) => Some(east, north, zoom, 10)
+      case (Some(east), Some(north), Some(zoom), Some(assetType)) => Some(east, north, zoom, assetType)
```
MAC-CGI: Since you are getting the assetTypeId in the previous line, this code could be replace by
```
val userPreferences: Option[(Long, Long, Int)] = (user.configuration.east, user.configuration.north, user.configuration.zoom) match {	  
  val assetTypeId = user.configuration.assetType match {
      case (Some(east), Some(north), Some(zoom)) => Some(east, north, zoom, assetTypeId )
```

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/MunicipalityDao.scala
```diff
@@ -82,4 +82,10 @@ class MunicipalityDao {
       sql"""select ELY_NRO from municipality  where id in (#${municipalities.mkString(",")} ) group by ELY_NRO""".as[Int].list
     }
   }
+
+  def getElysIdAndNamesByCode(elys: Set[Int]): Seq[(Int, String)] ={
```
MAC-CGI: delete this method. It is not needed

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -1871,13 +1876,56 @@ class Digiroad2Api(val roadLinkService: RoadLinkService,
     transformation(values)
   }
 
+  get("/userConfiguration/elys") {
+    val user = userProvider.getCurrentUser()
+    val elysId = municipalityDao.getElysByMunicipalities(user.configuration.authorizedMunicipalities)
+    val elysIdAndName = municipalityDao.getElysIdAndNamesByCode(elysId.toSet)
```
MAC-CGI: already exist a method which return MunicipalityInfo(id, ely, name) 

> getMunicipalitiesNameAndIdByCode

UI/src/view/defaultLocationButton.js
oskarmalmstrom: Maybe it is a part of a bigger issue and needs to be cleaned up, but in production there are users with ["premium", "busStopMaintainer"].

FilPinto: Ok i will rewrite this piece of code then

UI/src/view/initialPopupView.js
```diff
@@ -0,0 +1,215 @@
...
+        defaultParameters.lat = location.lat;
+        defaultParameters.zoom = location.zoom;
+      }
+      else if (selectedLocation === "suomi") {
+        defaultParameters.lon = 390000;
+        defaultParameters.lat = 6900000;
+        defaultParameters.zoom = 2;
+      }
+      else {
+        if (userRole !== "busStopMaintainer") {
```
FilPinto: Could use something like this `if (!isElyMaintainer())` by using the respective function in authorizationPolicy.js

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -212,34 +212,40 @@ class Digiroad2Api(val roadLinkService: RoadLinkService,
...
   def municipalityDao = new MunicipalityDao
 
-  private def getMapViewStartParameters(mapView: Option[MapViewZoom]): Option[(Double, Double, Int)] = mapView.map(m => (m.geometry.x, m.geometry.y, m.zoom))
+  private def getMapViewStartParameters(mapView: Option[MapViewZoom], assetType: Int): Option[(Double, Double, Int, Int)] = mapView.map(m => (m.geometry.x, m.geometry.y, m.zoom, assetType))
 
 
   get("/startupParameters") {
-    val defaultValues: (Double, Double, Int) = (390000, 6900000, 2)
+    val massTransitStopId = 10
+    val defaultValues: (Double, Double, Int, Int) = (390000, 6900000, 2, massTransitStopId)
```
FilPinto: You could replace `val massTransitStopId` with something like this `MassTransitStopAsset.typeId` since the asset.scala has the information, so no need to have that val

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
FilPinto: In this case I need the ely info (id and name) not the municipality info

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/MunicipalityDao.scala
FilPinto: After talking with the team the development is correct

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
FilPinto: After talking with the team the development is correct

PR #1431, opened 2019-09-24
Droth 2059 indicate conflicting field values within road link chain

marquesfm:


UI/src/view/link_property/linkPropertyForm.js
```diff
@@ -82,9 +82,23 @@
...
+        }
+        else{
+          verticalLevelType = _.find(verticalLevelTypes, function(y) { return y[0] === parseInt(verticalLevel, 10); });
+          return verticalLevelType && verticalLevelType[1];
+        }
+      }
+      else{
+        verticalLevelType = _.find(verticalLevelTypes, function(y) { return y[0] === verticalLevel; });
+        return verticalLevelType && verticalLevelType[1];
+      }
```
marquesfm: Can be more simplify like this:
```js
    var getVerticalLevelType = function(verticalLevel) {
      if (typeof verticalLevel === 'string') {
        var multipleLevels = verticalLevel.includes(",");
        if (multipleLevels) {
          return "[useita eri arvoja]";
        }
      }

      var verticalLevelType = _.find(verticalLevelTypes, function(y) { return y[0] === parseInt(verticalLevel, 10); });
      return verticalLevelType && verticalLevelType[1];
    };

PR #1435, opened 2019-10-15
Droth 2045 progress bar for oth map view

MAC-CGI:
Droth 2045 progress bar for oth map view


UI/src/utils/backend-utils.js
```diff
@@ -665,26 +665,26 @@
...
-                request.abort();
-
-            request = $.ajax(getParameters.apply(undefined, arguments)).done(function(result){
-                deferred.resolve(result);
-            });
-            return deferred;
-        }
+        request = $.ajax(getParameters.apply(undefined, arguments)).done(function (result) {
+          deferred.resolve(result);
+          eventbus.trigger('loadingBar:hide');
```
oskarmalmstrom: Should we hide the loading bar when the request is finished or when the assets are rendered on the map? If there are a lot of assets there can probably be some second of delay between the request and the rendering

UI/src/view/verificationWorkList.js
```diff
@@ -15,6 +15,14 @@
       });
     };
 
+    var addSpinner = function () {
```
ricardofmsimoes: Why not palce this functions 'addSpinner' and 'removeSpinner' in workListView.js, since this file inherits from workListView.js and we just can call this functions

UI/src/view/workListView.js
```diff
@@ -79,6 +79,14 @@
           .append(tableForGroupingValues('Ei tiedossa', workListItems.Unknown, 0));
     };
 
+    var addSpinner = function () {
+      $('#work-list').append('<div class="spinner-overlay modal-overlay"><div class="spinner"></div></div>');
```
ricardofmsimoes: Why not palce this functions 'addSpinner' and 'removeSpinner' in workListView.js, since this file inherits from workListView.js and we just can call this functions

UI/src/utils/backend-utils.js
FilPinto: > The only problem I see is the paralel calls like the manouvres/mass transit stops (these types of special cases) since one call will terminate and the other will not but will remove the loading bar nonetheless on the first call teminated
Fixed this one problem

PR #1439, opened 2019-11-11
DROTH-2171 Double checkmark on checked asset:

oskarmalmstrom:
-check for checkmark div, then append icon.
-changed the conditional for fetching roles on speed limit form. previous condition is always true if user is municipality user

### DIFF
--- a/UI/src/view/linear_asset/speedLimitForm.js
+++ b/UI/src/view/linear_asset/speedLimitForm.js
@@ -129,12 +129,8 @@
       }
   };
 
-  var bindEvents = function(selectedSpeedLimit, feedbackCollection, backend) {
+  var bindEvents = function(selectedSpeedLimit, feedbackCollection) {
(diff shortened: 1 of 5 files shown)


UI/src/view/linear_asset/speedLimitForm.js
```diff
@@ -132,7 +132,7 @@
   var bindEvents = function(selectedSpeedLimit, feedbackCollection, backend) {
     new FeedbackDataTool(feedbackCollection, 'speedLimit', authorizationPolicy);
 
-    if (authorizationPolicy.userRoles.length === 0){
+    if (!authorizationPolicy.fetched){
```
MAC-CGI: This code need to be re-think, Doesn't seem correct start call the getUserRoles in differents part of application. Could you investigate if it possible use the eventbus.on(roles:fecthed) and remove this?

PR #1443, opened 2019-11-27
DROTH-2189 fixes on newuser page

oskarmalmstrom:
-selected municipalities persisted after checkbox was unchecked, now they are removed.
-a 200-response is handled as an error. probably because we are expecting a json-response but not getting one because the request is sent as a serialized form.


UI/newuser.html
```diff
@@ -181,9 +184,12 @@
                     func(username, successResponseValue);
                 },
                 error: function (errorValue) {
-                    if (errorValue.status === 400) {
+                    if (errorValue.status >= 400) {
                         alert('Tarkista syöttämäsi tiedot.');
+                    } else if (errorValue.status === 200) {
+                        func(username, successResponseValue);
```
FilPinto: Can this be possible?
If the status code is 200 it will go to success function above and never enter this one, right?

oskarmalmstrom: from what i understand success/error isn't only defined by the status. The response also needs to conform with the dataType of the request. This is not defined because we are sending form data, so setting it breaks the back end. Maybe there is some serialization that would allow us to set the data type, couldn't find it directly. If someone knows, please change it!

FilPinto: From what I read it is possible to go to the error function with status code 200 IF ajax can't parse the response(for example, can be expecting a json and the server sended a XML). Since we are always expecting a json response it would be impossible, I think, to enter this part of the code.

PR #1451, opened 2019-12-18
Droth 2172 traffic control devices for pedestrian and cycleways

oskarmalmstrom:


UI/src/view/point_asset/trafficSignForm.js
```diff
@@ -75,6 +75,11 @@
 
     this.boxEvents = function(rootElement, selectedAsset, localizedTexts, authorizationPolicy, roadCollection, collection) {
 
+      rootElement.find('.form-traffic-sign').on('change', function(event) {
+          // force the field to be filled
+          selectedAsset.setPropertyByPublicId('opposite_side_sign', '0');
+      });
+
```
oskarmalmstrom: Does this need to be forced? I don't think it should be mandatory.

UI/src/view/point_asset/trafficLayer.js
```diff
@@ -0,0 +1,53 @@
...
+
+         this.handleMapClick = function (coordinates) {
+            if (application.getSelectedTool() === 'Add' && zoomlevels.isInAssetZoomLevel(zoomlevels.getViewZoom(map))) {
+                me.createNewAsset(coordinates);
+            } else if (selectedAsset.isDirty()) {
+                me.displayConfirmMessage();
+            }
+        };
+
+        this.createNewAsset = function(coordinates) {
```
oskarmalmstrom: The only difference with this function and the parent one is a different road collection parameter for excluding road links. It could probably be shortened in some way.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/RoadLinkService.scala
```diff
@@ -925,21 +925,21 @@ class RoadLinkService(val vvhClient: VVHClient, val eventbus: DigiroadEventBus,
...
+    }
 
-    roadLinks.isEmpty match {
-      case true => Seq.empty[VVHRoadlink]
-      case false => roadLinks.filter(rl => GeometryUtils.minimumDistance(point, rl.geometry) <= 10.0).filter(_.featureClass != FeatureClass.CycleOrPedestrianPath)
+    val closestRoadLinks = roadLinks.isEmpty match {
+      case true => Seq.empty[RoadLink]
+      case false => enrichRoadLinksFromVVH(roadLinks.filter(rl => GeometryUtils.minimumDistance(point, rl.geometry) <= 10.0))
     }
+    if (forCarTraffic) closestRoadLinks.filter(_.linkType != CycleOrPedestrianPath ) else closestRoadLinks
```
oskarmalmstrom: I think this filtering could be done where it's called, in this case in verifyData

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/TrafficSign.scala
```diff
@@ -30,6 +31,9 @@ sealed trait TrafficSignType {
 
   def group: TrafficSignTypeGroup
 
+  //This is only used to put CycleAndWalkwaySigns group at the moment
+  def additionalGroup: Option[TrafficSignTypeGroup] = None
+
```
FilPinto: If we start to having more signs with "mixed" groups maybe we should rethink the `def group: TrafficSignTypeGroup` to `def group: Seq[TrafficSignTypeGroup]` or something like that.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataImporter/MassTransitStopCsvImporter.scala
```diff
@@ -426,7 +423,7 @@ class PositionUpdater (vvhClientImpl: VVHClient, roadLinkServiceImpl: RoadLinkSe
     val lat = getPropertyValue(properties, "maastokoordinaatti_y").asInstanceOf[BigDecimal].toDouble
     val externalId = getPropertyValue(properties, "external_id").toString.toInt
 
-    val roadLink = roadLinkService.enrichRoadLinksFromVVH(getNearestRoadLink(lon, lat, user, roadTypeLimitations))
```
FilPinto: Why delete the enrich? We should update the roadlink we get from getNearestRoadLink method with some of our values in database right?

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/RoadLinkService.scala
FilPinto: I think we should use the method above getClosestRoadlinkFromVVH and change the `val diagonal = Vector3d(500, 500, 0)` to receive a parameter with a default value 500. Since the idea here when putting the forCarTraffic argument to false is the same as using the getClosestRoadlinkFromVVH  method.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataImporter/MassTransitStopCsvImporter.scala
oskarmalmstrom: The enrich is done on `def getClosestRoadlinkForCarTrafficFromVVH` on RoadLinkService.scala

UI/src/view/point_asset/trafficLayer.js
```diff
@@ -0,0 +1,53 @@
+(function(root) {
+    root.TrafficLayer = function(params) {
```
oskarmalmstrom: I think TrafficSignLayer is a better name

UI/src/view/point_asset/trafficSignForm.js
ricardofmsimoes: Yes, this field need to be forced.
Otherwise we can't update the traffic signs with a single choice with value ""

UI/src/view/point_asset/trafficLayer.js
RobertVisan: done

PR #1452, opened 2019-12-19
DROTH-2216 Obstacle batch

FilPinto:
-Created new batch to add obstacles from the shapefile
-Created query to fetch the shapefile info and transform it to Obstacle
-Added the batch to the task runners
-Modify the getClosestRoadlinkFromVVH method to receive a radius

### DIFF
--- a/digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
+++ b/digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
@@ -13,7 +13,7 @@ import fi.liikennevirasto.digiroad2.client.vvh.VVHClient
 import fi.liikennevirasto.digiroad2.dao._
 import fi.liikennevirasto.digiroad2.dao.linearasset.{OracleLinearAssetDao, OracleSpeedLimitDao}
 import fi.liikennevirasto.digiroad2.linearasset._
-import fi.liikennevirasto.digiroad2.dao.pointasset.Obstacle
+import fi.liikennevirasto.digiroad2.dao.pointasset.{Obstacle, ObstacleShapefile}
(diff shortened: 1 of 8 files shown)


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -1818,6 +1818,53 @@ object DataFixture {
...
+      obstacleService.getObstaclesFromShapefileTable()
+    }
+
+    OracleDatabase.withDynTransaction {
+      obstaclesInformation.foreach { obstacle =>
+        println("")
+        println("Creating a obstacle with coordinates -> " + "x:" + obstacle.lon + " y:" + obstacle.lat)
+        val pointObstacle = Point(obstacle.lon, obstacle.lat)
+
+        roadLinkService.getClosestRoadlinkFromVVH(user, pointObstacle, 10) match {
```
marquesfm: Normally in these cases, it's better to perform only one request to VVH instead of one for each obstacle. 
But if this batch it's to be run only one-time let's leave like it is.

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -1818,6 +1818,53 @@ object DataFixture {
...
+          case Some(link) =>
+            val nearestRoadLinks = roadLinkService.enrichRoadLinksFromVVH(Seq(link))
+
+            if(nearestRoadLinks.nonEmpty){
+              val nearestRoadLink = nearestRoadLinks.head
```
marquesfm: Doesn't have a rule in case of more than one nearest road link?
Does Mikko say something about that?

FilPinto: It will only get one because when getting from VVH you only bring the closer one

PR #1456, opened 2020-01-20
Hotfix jan 2020 qa

ricardofmsimoes:


UI/src/assetTypeConfiguration.js
```diff
@@ -988,6 +989,11 @@
             { types: [8, 30, 31, 32, 33, 34, 35], validate: function (someValue) { return /^\d*\.?\d+$/.test(someValue) ; }}
           ];
 
+          var opposite_side_sign =  selectedAsset.get().propertyData.find( function(prop) { if (prop.publicId === "opposite_side_sign") return prop; }) ;
```
marquesfm:
```scala
var opposite_side_sign =  _.find(selectedAsset.get().propertyData, function(prop){ if (prop.publicId === "opposite_side_sign") return prop; }) ;

File: digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataImporter/MassTransitStopCsvImporter.scala [hunk 2]
```diff
@@ -42,9 +44,17 @@ class MassTransitStopCsvOperation(vvhClientImpl: VVHClient, roadLinkServiceImpl:
...
+      val strategy = getStrategy(csvRow.head)
+      strategy.importAssets(csvRow: List[Map[String, String]], fileName, user, logId, roadTypeLimitations)
+
+    } catch {
+      case e: Exception =>
+        OracleDatabase.withDynTransaction  {
+        importLogDao.update(logId, Status.Abend, Some("Latauksessa tapahtui odottamaton virhe: " + e.toString))
+        }
+      }
+  }
```
FilPinto: This try catch is not being done in line 282?

marquesfm: This is necessary because when the strategy fails it's being generated a throw exception not caught and the error is not logged.

PR #1459, opened 2020-01-24
split changeApi into paging

MAC-CGI:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/linearasset/OracleSpeedLimitDao.scala
```diff
@@ -293,30 +294,42 @@ class OracleSpeedLimitDao(val vvhClient: VVHClient, val roadLinkService: RoadLin
     (speedLimitLinks, roadLinks)
   }
 
-  def getSpeedLimitsChangedSince(sinceDate: DateTime, untilDate: DateTime, withAdjust: Boolean): Seq[PersistedSpeedLimit] = {
+  def getSpeedLimitsChangedSince(sinceDate: DateTime, untilDate: DateTime, withAdjust: Boolean, pageNumber: Option[Int]): Seq[PersistedSpeedLimit] = {
     val withAutoAdjustFilter = if (withAdjust) "" else "and (a.modified_by is null OR a.modified_by != 'vvh_generated')"
+    val recordLimit = pageNumber match {
```
ricardofmsimoes: Since this code is used in multiple location maybe it should good to have it in query file or other file like:

def pageNumberCalc ( pageNumber: Option[Int] )
pageNumber match {
      case Some(pgNum) =>
        val startNum = (RECORD_NUMBER) * (pgNum - 1) + 1
        val endNum = pgNum * RECORD_NUMBER
        s"WHERE line_number between $startNum and $endNum"
      case _ => ""
    }
}

PR #1463, opened 2020-02-02
DROTH-2243 Asset deletes when geometry updates

MAC-CGI:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/linearasset/OracleLinearAssetDao.scala
```diff
@@ -1006,4 +1006,17 @@ class OracleLinearAssetDao(val vvhClient: VVHClient, val roadLinkService: RoadLi
          and ca.created_date > ADD_MONTHS(TO_DATE(TO_CHAR(${lastCreationDate.get}, 'YYYY-MM-DD'), 'YYYY-MM-DD hh24:mi:ss'), -1)
          #$municipalityFilter""".as[(Long, Int)].list
   }
+
+  def getLinksWithExpiredAssets(linkIds: Seq[Long], assetType: Int): Seq[Long] = {
+    val withLinkIds = if (linkIds.nonEmpty) s"and pos.link_id in (${linkIds.mkString(",")})" else ""
```
ricardofmsimoes: shouldn't this IN be replaced with MassQuery ?

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/linearasset/LinearAssetService.scala
```diff
@@ -306,13 +305,73 @@ trait LinearAssetOperations {
...
-    val linearAssets = mapReplacementProjections(assetsToUpdate, currentAssets, roadLinks, fullChanges).foldLeft((Seq.empty[PersistedLinearAsset], changeSet)) {
-      case ((persistedAsset, cs), (asset, (Some(roadLink), Some(projection)))) =>
-        val (linearAsset, changes) = assetFiller.projectLinearAsset(asset, roadLink, projection, cs)
-        (persistedAsset ++ Seq(linearAsset), changes)
-      case _ => (Seq.empty[PersistedLinearAsset], changeSet)
+
+    val linearAssetsAndChanges = mapReplacementProjections(assetsToUpdate, currentAssets, roadLinks, fullChanges).flatMap {
+      case (asset, (Some(roadLink), Some(projection))) =>
+        val (linearAsset, changes) = assetFiller.projectLinearAsset(asset, roadLink, projection, changeSet)
+        Some((linearAsset, changes))
```
marquesfm: I think can be switch to 
```scala
assetFiller.projectLinearAsset(asset, roadLink, projection, changeSet).toOption

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/linearasset/OracleLinearAssetDao.scala
marquesfm: In this case, we use the IN clause because it's supposed to return even if the link ids Sequence was empty, and if we do that using the mass query we never get results.

PR #1468, opened 2020-02-06
Droth 2274 split change api queries into pages v2

MAC-CGI:


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/DynamicLinearAssetDao.scala
```diff
@@ -430,8 +433,18 @@ class DynamicLinearAssetDao {
...
   }
 
-  def getDynamicLinearAssetsChangedSince(assetTypeId: Int, sinceDate: DateTime, untilDate: DateTime, withAdjust: Boolean, recordLimit: String) : List[PersistedLinearAsset] = {
+  def getDynamicLinearAssetsChangedSince(assetTypeId: Int, sinceDate: DateTime, untilDate: DateTime, withAdjust: Boolean, token: Option[String] = None) : List[PersistedLinearAsset] = {
     val withAutoAdjustFilter = if (withAdjust) "" else "and (a.modified_by is null OR a.modified_by != 'vvh_generated')"
+    val recordLimit = token match {
+      case Some(tk) =>
+        val values = Decode.getPageAndRecordNumber(tk)
+
+        val startNum =values("recordNumber") * ( values("pageNumber")- 1) + 1
```
ricardofmsimoes: This parte of the code is repeted in other files
` val values = Decode.getPageAndRecordNumber(tk)

        val startNum =values("recordNumber") * ( values("pageNumber")- 1) + 1
        val endNum = values("pageNumber") * values("recordNumber")
        s"WHERE line_number between $startNum and $endNum"`

Can't it be pushed to some auxiliar function?

marquesfm: I agree, let's bring startNum and endNum calculation to inside getPageAndRecordNumber method like:
```scala
  def getPageAndRecordNumber(src: String): (Int, Int) = {
    val values = new String(Base64.getDecoder.decode(src), StandardCharsets.UTF_8).split(',')
      .flatMap(_.split(':').map(_.trim) match {
        case Array(s: String, i: String) => Map(s -> i.toInt)
      }).toMap

    val startNum = values("recordNumber") * (values("pageNumber") - 1) + 1
    val endNum = values("pageNumber") * values("recordNumber")

    (startNum, endNum)
  }
```
And then on each use modify like: 
```scala
val (startNum, endNum) = Decode.getPageAndRecordNumber(tk)

PR #1469, opened 2020-02-11
DROTH-2273 (US 1891) Suggested assets-button is missing in municipali…

FilPinto:
…ty management page

-Fixed problem when clicking suggested asset button and it gave an error in server (500)
-Fixed expired suggested assets appearing on the list
-Fixed problem when clicking suggested asset button and no assets are shown

### DIFF
--- a/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/VerificationDao.scala
+++ b/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/VerificationDao.scala
@@ -39,7 +39,7 @@ class VerificationDao {
         from municipality_verification mv
         join asset_type atype on atype.id = mv.asset_type_id
         join municipality m on m.id = mv.municipality_id
-        where m.id = $municipalityId and atype.id = $typeId
+        where m.id = $municipalityId and atype.id = $typeId and valid_to is null
       """.as[(String, Int, String, String)].list
     suggestedAssets.map { row =>
       SuggestedAssetsStructure(row._1, row._2, row._3, typeId, row._4.split(",").map(_.toInt).toSeq)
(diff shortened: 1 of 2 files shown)


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/VerificationDao.scala
```diff
@@ -115,6 +115,7 @@ class VerificationDao {
         join property p on p.asset_type_id = a.asset_type_id and p.property_type = 'checkbox' and p.public_id = 'suggest_box'
         join multiple_choice_value mcv on mcv.asset_id = a.id and mcv.property_id = p.id
         join enumerated_value ev on ev.property_id = p.id and ev.value = 1 and ev.id = mcv.enumerated_value_id and  lrm.link_id in (select id from #$idTableName)
```
ricardofmsimoes: This part " and  lrm.link_id in (select id from #$idTableName)"
Shouldn't be at WHERE clause ?

FilPinto: I left it as it was. No problem when executing the query.

marquesfm: Why not at line 114?
`scala

join lrm_position lrm on lrm.id = al.position_id and lrm.link_id in (select id from #$idTableName)
`

marquesfm: Let's keep it like it is, it's working and already has reverted to different ways two times and the final result was always this.

PR #1476, opened 2020-03-05
Droth 2306 some bugs in unknown speed limit list

FilPinto:


digiroad2-api-common/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Context.scala
```diff
@@ -165,7 +165,7 @@ class DynamicAssetSaveProjected[T](dynamicAssetProvider: DynamicLinearAssetServi
 
 class SpeedLimitUpdater[A, B, C](speedLimitProvider: SpeedLimitService) extends Actor {
   def receive = {
-    case x: Set[A] => speedLimitProvider.purgeUnknown(x.asInstanceOf[Set[Long]])
+    case x: (Set[A], Seq[A]) => speedLimitProvider.purgeUnknown(x._1.asInstanceOf[Set[Long]], x._2.asInstanceOf[Seq[Long]])
```
ricardofmsimoes: Sugestion:

```scala
 case (x, y): (Set[A], Seq[A]) => speedLimitProvider.purgeUnknown(x.asInstanceOf[Set[Long]], y.asInstanceOf[Seq[Long]]) ```

FilPinto: Maybe it's better to give good names to the variables to be more readable

FilPinto: Applied both ideas

PR #1479, opened 2020-03-10
Droth 2307 oth crashes on production servers

marquesfm:


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -2306,6 +2307,56 @@ object DataFixture {
...
+    }
+
+    OracleDatabase.withDynTransaction {
+      municipalities.foreach { municipality =>
+        println("\nWorking at Municipailty: " + municipality)
+        val (roadLinks, changes) = roadLinkService.getRoadLinksWithComplementaryAndChangesFromVVHByMunicipality(municipality, newTransaction = false)
+        val filteredRoadLinks = roadLinks.filter(r => r.isCarRoadOrCyclePedestrianPath)
+        val changesToTreat = changes.filter(c => c.changeType == New.value && filteredRoadLinks.exists(_.linkId == c.newId.getOrElse(0)))
+        val speedLimitsAlreadyExistents = changesToTreat.flatMap { ctt =>
+          speedLimitService.getExistingAssetByRoadLink(filteredRoadLinks.find(_.linkId == ctt.newId.getOrElse(0)).get, false)
```
FilPinto: I think there is no need to do a getOrElse in these two scenarios because in the first condition you already confirm that it is a New changeType then you are sure that it has a newId so a simple get would be sufficient.
And line 2333.

FilPinto: In line 2327 aren't you doing too many calls to db?
I think would be better like this:
```scala
val roadLinksToTreat = filteredRoadLinks.filter(r => changesToTreat.exists(_.newId.get == r.linkId))
val speedLimitsAlreadyExistents = speedLimitService.getExistingAssetByRoadLink(roadLinksToTreat, false)
```

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -2306,6 +2307,56 @@ object DataFixture {
...
+        }
+        val changesWithoutSpeedLimitCreated = changesToTreat.filterNot(ctt => speedLimitsAlreadyExistents.exists(_.linkId == ctt.newId.get))
+        println("\nNumber of Changes to Treat: "+ changesWithoutSpeedLimitCreated)
+
+        changesWithoutSpeedLimitCreated.foreach { cws =>
+          val adjacents = roadLinkService.getAdjacent(cws.newId.getOrElse(0L), false).map(_.linkId).toSet
+          if (adjacents.size >= 2) {
+            val speedLimitsOnAdjacents = speedLimitDao.getCurrentSpeedLimitsByLinkIds(Some(adjacents))
+            if (speedLimitsOnAdjacents.size >= 2) {
+              val speedLimitsToCreate = speedLimitService.newChangeAsset(filteredRoadLinks, speedLimitsOnAdjacents, Seq(cws))
```
FilPinto: You can use the val roadLinksToTreat in my previous suggestion to replace the filteredRoadLinks. It will contain less roadLinks so the find method inside this method will take less time

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -2306,6 +2307,56 @@ object DataFixture {
...
+        val speedLimitsAlreadyExistents = changesToTreat.flatMap { ctt =>
+          speedLimitService.getExistingAssetByRoadLink(filteredRoadLinks.find(_.linkId == ctt.newId.getOrElse(0)).get, false)
+        }
+        val changesWithoutSpeedLimitCreated = changesToTreat.filterNot(ctt => speedLimitsAlreadyExistents.exists(_.linkId == ctt.newId.get))
+        println("\nNumber of Changes to Treat: "+ changesWithoutSpeedLimitCreated)
+
+        changesWithoutSpeedLimitCreated.foreach { cws =>
+          val adjacents = roadLinkService.getAdjacent(cws.newId.getOrElse(0L), false).map(_.linkId).toSet
+          if (adjacents.size >= 2) {
+            val speedLimitsOnAdjacents = speedLimitDao.getCurrentSpeedLimitsByLinkIds(Some(adjacents))
```
FilPinto: You already have the val speedLimitsAlreadyExistents I think you can use it instead of going to the database various times

```scala
val speedLimitsOnAdjacents = speedLimitsAlreadyExistents.filter(s => adjacents.exist(s.linkId))
```

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -2306,6 +2307,56 @@ object DataFixture {
...
+    val municipalities: Seq[Int] = OracleDatabase.withDynSession {
+      Queries.getMunicipalities
+    }
+
+    OracleDatabase.withDynTransaction {
+      municipalities.foreach { municipality =>
+        println("\nWorking at Municipailty: " + municipality)
+        val (roadLinks, changes) = roadLinkService.getRoadLinksWithComplementaryAndChangesFromVVHByMunicipality(municipality, newTransaction = false)
+        val filteredRoadLinks = roadLinks.filter(r => r.isCarRoadOrCyclePedestrianPath)
+        val changesToTreat = changes.filter(c => c.changeType == New.value && filteredRoadLinks.exists(_.linkId == c.newId.getOrElse(0)))
```
MAC-CGI: could be change like that `changes.filter(c => c.changeType == New.value && c.newId.nonEmpty &&  filteredRoadLinks.exists(_.linkId == c.newId.get))`

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -2306,6 +2307,56 @@ object DataFixture {
...
+        val speedLimitsAlreadyExistents = changesToTreat.flatMap { ctt =>
+          speedLimitService.getExistingAssetByRoadLink(filteredRoadLinks.find(_.linkId == ctt.newId.getOrElse(0)).get, false)
+        }
+        val changesWithoutSpeedLimitCreated = changesToTreat.filterNot(ctt => speedLimitsAlreadyExistents.exists(_.linkId == ctt.newId.get))
+        println("\nNumber of Changes to Treat: "+ changesWithoutSpeedLimitCreated)
+
+        changesWithoutSpeedLimitCreated.foreach { cws =>
+          val adjacents = roadLinkService.getAdjacent(cws.newId.getOrElse(0L), false).map(_.linkId).toSet
```
MAC-CGI: with previous approach this could be replace like 
` val adjacents = roadLinkService.getAdjacent(cws.newId.get, false).map(_.linkId).toSet`

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
FilPinto: Sorry this should not be applied

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
marquesfm: This not make sense, since with this you will have only the speed limits in the changes already treated, and we want the speed limits in the adjacents road links, if they exist.

PR #1481, opened 2020-03-11
DROTH-2307 OTH crashes on both production servers

marquesfm:
- Added code to contemplate road without exit but with previus speedlimit

### DIFF
--- a/digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
+++ b/digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
@@ -2319,19 +2319,19 @@ object DataFixture {
         println("New SpeedLimit created at Link Id: " + speedLimit.linkId + " with value: " + speedLimit.value.get.value + " and sidecode: " + speedLimit.sideCode)
 
         //Remove linkIds from Unknown Speed Limits working list after speedLimit creation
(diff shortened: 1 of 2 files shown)


digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/util/DataFixture.scala
```diff
@@ -2308,6 +2308,22 @@ object DataFixture {
...
+    }
+
+    def createNewSpeedLimits(newSpeedLimits: Seq[SpeedLimit]): Unit = {
+      //Create new SpeedLimits on gaps
+      newSpeedLimits.foreach { speedLimit =>
+        speedLimitDao.createSpeedLimit(LinearAssetTypes.VvhGenerated, speedLimit.linkId, Measures(speedLimit.startMeasure, speedLimit.endMeasure), speedLimit.sideCode, speedLimit.value.get, vvhClient.roadLinkData.createVVHTimeStamp(), (_, _) => Unit)
+        println("New SpeedLimit created at Link Id: " + speedLimit.linkId + " with value: " + speedLimit.value.get.value)
+
+        //Remove linkIds from Unknown Speed Limits working list after speedLimit creation
+        speedLimitDao.purgeFromUnknownSpeedLimits(speedLimit.linkId, GeometryUtils.geometryLength(speedLimit.geometry))
```
MAC-CGI: check if we can thrust on speedLimit.geometry, probably we should use roadlink.geometry

PR #1482, opened 2020-03-16
DROTH-2268 Bus stops that need to be deleted

FilPinto:
-Override def isFloating in TerminatedBusStopStrategy so it will not go to the parent to give the wrong boolean and reason


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/pointasset/masstransitstop/TerminatedBusStopStrategy.scala
```diff
@@ -51,4 +51,5 @@ class TerminatedBusStopStrategy(typeId: Int, massTransitStopDao: MassTransitStop
     (resultAsset, PublishInfo(Some(resultAsset)))
   }
 
+  override def isFloating(persistedAsset: PersistedMassTransitStop, roadLinkOption: Option[RoadLinkLike]): (Boolean, Option[FloatingReason]) = { (true, Some(FloatingReason.TerminatedRoad)) }
```
FilPinto: The isFloating from TierekisteriBusStopStrategy is doing something similar, if not the same, to this that I think that method is not doing anything, only returning (false, None).

PR #1484, opened 2020-03-17
DROTH-2321Modifing rights to state roads in Ahvenanmaa

ricardofmsimoes:
- Added a modification rights to state roads in Ahvenanmaa

### DIFF
--- a/UI/src/authorizationpolicies/authorizationPolicy.js
+++ b/UI/src/authorizationpolicies/authorizationPolicy.js
@@ -79,11 +79,11 @@
       return (selectedAsset.isNew() && me.isOperator()) || (suggestedBoxValue && (me.isOperator() || me.isMunicipalityMaintainer()));
     };
 
-    this.isMunicipalityExcluded = function (selectedAsset) {
-      var municipalitiesExcluded = [478,60,65,76,170,736,771,43,417,438,35,62,295,318,766,941];
+    this.isStateExclusions = function (selectedAsset) {
+      var statesExcluded = [35,43,60,62,65,76,170,295,318,417,438,478,736,766,771,941];
       var isOperatorAndHaveRights = me.isOperator() && me.hasRightsInMunicipality(selectedAsset.municipalityCode);
 
-      return municipalitiesExcluded.indexOf(selectedAsset.municipalityCode) >= 0 && isOperatorAndHaveRights;
+      return _.includes(statesExcluded, selectedAsset.municipalityCode) && isOperatorAndHaveRights;
     };
 
   };
--- a/UI/src/authorizationpolicies/authorizationPolicy.js
+++ b/UI/src/authorizationpolicies/authorizationPolicy.js
@@ -79,5 +79,12 @@
       return (selectedAsset.isNew() && me.isOperator()) || (suggestedBoxValue && (me.isOperator() || me.isMunicipalityMaintainer()));
     };
 
+    this.isMunicipalityExcluded = function (selectedAsset) {
+      var municipalitiesExcluded = [478,60,65,76,170,736,771,43,417,438,35,62,295,318,766,941];
+      var isOperatorAndHaveRights = me.isOperator() && me.hasRightsInMunicipality(selectedAsset.municipalityCode);
+
+      return municipalitiesExcluded.indexOf(selectedAsset.municipalityCode) >= 0 && isOperatorAndHaveRights;
+    };
+
   };
 })(this);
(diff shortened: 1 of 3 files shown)


UI/src/authorizationpolicies/authorizationPolicy.js
```diff
@@ -79,5 +79,12 @@
       return (selectedAsset.isNew() && me.isOperator()) || (suggestedBoxValue && (me.isOperator() || me.isMunicipalityMaintainer()));
     };
 
+    this.isMunicipalityExcluded = function (selectedAsset) {
+      var municipalitiesExcluded = [478,60,65,76,170,736,771,43,417,438,35,62,295,318,766,941];
+      var isOperatorAndHaveRights = me.isOperator() && me.hasRightsInMunicipality(selectedAsset.municipalityCode);
+
+      return municipalitiesExcluded.indexOf(selectedAsset.municipalityCode) >= 0 && isOperatorAndHaveRights;
```
FilPinto: I think it will look better if we use lodash in this to avoid that ">=0"
```javascript
return _.includes(municipalitiesExcluded, selectedAsset.municipalityCode) && isOperatorAndHaveRights;
```

UI/src/authorizationpolicies/authorizationPolicy.js
```diff
@@ -79,5 +79,12 @@
       return (selectedAsset.isNew() && me.isOperator()) || (suggestedBoxValue && (me.isOperator() || me.isMunicipalityMaintainer()));
     };
 
+    this.isMunicipalityExcluded = function (selectedAsset) {
+      var municipalitiesExcluded = [478,60,65,76,170,736,771,43,417,438,35,62,295,318,766,941];
```
marquesfm: Can we put some order on this?

UI/src/authorizationpolicies/linearAssetAuthorizationPolicy.js
```diff
@@ -5,7 +5,10 @@
     var me = this;
 
     this.formEditModeAccess = function(selectedAsset) {
-      return (me.isMunicipalityMaintainer() && !me.isState(selectedAsset) && me.hasRightsInMunicipality(selectedAsset.municipalityCode)) || (me.isElyMaintainer() && me.hasRightsInMunicipality(selectedAsset.municipalityCode)) || me.isOperator();
+      var isMunicipalityMaintainerAndHaveRights = me.isMunicipalityMaintainer() && !me.isState(selectedAsset) && me.hasRightsInMunicipality(selectedAsset.municipalityCode);
+      var isElyMaintainerAndHasRights = me.isElyMaintainer() && me.hasRightsInMunicipality(selectedAsset.municipalityCode);
+
+      return  me.isMunicipalityExcluded(selectedAsset) ||  ( isMunicipalityMaintainerAndHaveRights || isElyMaintainerAndHasRights || me.isOperator() );
```
marquesfm: It's really needed this modification?

It's because this method it's overrided at linearStateRoadAuthorizationPolicy.js file so I think the only modification needed it's the applied bellow.

Because with that they will apply these rules in all linear assets and we probably don't want that.

UI/src/authorizationpolicies/authorizationPolicy.js
```diff
@@ -79,5 +79,12 @@
       return (selectedAsset.isNew() && me.isOperator()) || (suggestedBoxValue && (me.isOperator() || me.isMunicipalityMaintainer()));
     };
 
+    this.isMunicipalityExcluded = function (selectedAsset) {
```
marquesfm: Let's put a better name, for example, something like isStateExclusions or something like that.

UI/src/authorizationpolicies/linearAssetAuthorizationPolicy.js
ricardofmsimoes: lengthLimit and widthLimit uses LinearAssetAuthorizationPolicyWith Suggestion and this object calls linearAsset AuthorizationPolicy to validate if a form should or not be able to edit

UI/src/authorizationpolicies/authorizationPolicy.js
ricardofmsimoes: Done.

UI/src/authorizationpolicies/authorizationPolicy.js
ricardofmsimoes: Done.

UI/src/authorizationpolicies/authorizationPolicy.js
ricardofmsimoes: Done.

PR #1490, opened 2020-03-24
Droth 2335 extraction railway and pedestrian crossing fails

ricardofmsimoes:
-Fix som extrations


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/pointasset/PropertyValidator.scala
```diff
@@ -0,0 +1,20 @@
+package fi.liikennevirasto.digiroad2.dao.pointasset
+
+object PropertyValidator {
+
+  private val DEFAULT_PROPERTIES_VALUES = Map("suggest_box" -> "0",
```
marquesfm: Modify the name of this to "defaultPropertiesValues"

PR #1492, opened 2020-03-24
DROTH-2332 Service road problems

RobertVisan:
-Fixed problems on dynamic linear asset selection that would lock the user on a linear asset when selected
-Fixed visual bug on checkbox view mode

### DIFF
--- a/UI/src/assetTypeConfiguration.js
+++ b/UI/src/assetTypeConfiguration.js
@@ -125,7 +125,7 @@
         form: new DynamicAssetForm({
           fields: [
             {label: "massarajoitus", type: 'integer', publicId: "weight", unit: "Kg", required: true, weight: 1},
-            {label: "vihjetieto", type: 'checkbox', publicId: "suggest_box", weight: 2, showAndHide: showSuggestBox, isUnSet: isSuggestBoxUnset}
+            {label: "vihjetieto", type: 'checkbox', publicId: "suggest_box", defaultValue: "0", values: [{id: 0, label: 'Tarkistettu'}, {id: 1, label: 'Vihjetieto'}], weight: 2, showAndHide: showSuggestBox, isUnSet: isSuggestBoxUnset}
           ]
         })
       },
@@ -157,7 +157,7 @@
         form: new DynamicAssetForm({
           fields: [
             {label: "massarajoitus", type: 'integer', publicId: "weight", unit: "Kg", required: true, weight: 1},
-            {label: "vihjetieto", type: 'checkbox', publicId: "suggest_box", weight: 2, showAndHide: showSuggestBox, isUnSet: isSuggestBoxUnset}
+            {label: "vihjetieto", type: 'checkbox', publicId: "suggest_box", defaultValue: "0", values: [{id: 0, label: 'Tarkistettu'}, {id: 1, label: 'Vihjetieto'}], weight: 2, showAndHide: showSuggestBox, isUnSet: isSuggestBoxUnset}
           ]
         })
       },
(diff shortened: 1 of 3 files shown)


UI/src/view/dynamicAssetForm.js
```diff
@@ -85,13 +85,13 @@
...
 
-        me.setSelectedValue = function(setValue, getValue){
+        me.setSelectedValue = function(setValue, getValue, silently) {
 
             var currentPropertyValue = me.hasValue() ?  me.getPropertyValue() : (me.hasDefaultValue() ? me.getPropertyDefaultValue() : me.emptyPropertyValue());
 
             var properties = _.filter(getValue() ? getValue().properties : getValue(), function(property){ return property.publicId !== currentPropertyValue.publicId; });
             var value = properties.concat(currentPropertyValue);
-            setValue({ properties: value});
+            setValue({ properties: value}, silently);
```
FilPinto: This setValue can be the selectedLinearAsset.setValue() as you are seeing but it can be the setAValue or setBValue also so you will probably need to change those too, check please

UI/src/model/selectedLinearAsset.js
```diff
@@ -280,12 +280,14 @@
       return selection.length;
     };
 
-    this.setValue = function(value) {
+    this.setValue = function(value, silently) {
```
FilPinto: I don't understand what this "silently" is can you change into another name to be more recognizable

RobertVisan: I think silently is a good name because you want to make changes to the selected asset without firing any events

UI/src/view/dynamicAssetForm.js
RobertVisan: I will check, this is mainly used when affecting a property with some default value if this is not done the behaviour should stay the same

UI/src/assetTypeConfiguration.js
```diff
@@ -125,7 +125,7 @@
         form: new DynamicAssetForm({
           fields: [
             {label: "massarajoitus", type: 'integer', publicId: "weight", unit: "Kg", required: true, weight: 1},
-            {label: "vihjetieto", type: 'checkbox', publicId: "suggest_box", weight: 2, showAndHide: showSuggestBox, isUnSet: isSuggestBoxUnset}
+            {label: "vihjetieto", type: 'checkbox', publicId: "suggest_box", defaultValue: "0", values: [{id: 0, label: 'Tarkistettu'}, {id: 1, label: 'Vihjetieto'}], weight: 2, showAndHide: showSuggestBox, isUnSet: isSuggestBoxUnset}
```
marquesfm: replace id for value here and in similars
```javascript
values: [{value: 0, label: 'Tarkistettu'}, {value: 1, label: 'Vihjetieto'}]

UI/src/model/selectedLinearAsset.js
marquesfm: can be replaced to "isToFireEvents" for example??

RobertVisan: changed it to withoutEventTriggers

UI/src/assetTypeConfiguration.js
RobertVisan: Changing this would imply that I would have to change all the fields for all the other assets in order to access them by the value key instead of the id key, not only the suggest_box

UI/src/view/dynamicAssetForm.js
RobertVisan: setAValue and setBValue is used for creating two new linear assets and on those cases you want to trigger events so it won't be needed here

PR #1493, opened 2020-03-27
DROTH-2344 (US 2142) Value field disappears

FilPinto:
-Fixed name/weight fields not being saved as they should

### DIFF
--- a/UI/src/view/point_asset/servicePointForm.js
+++ b/UI/src/view/point_asset/servicePointForm.js
@@ -251,14 +251,14 @@
         delete service.weightLimit;
     }
 
-    function checkNimiField(service, modifications)  {
+    function checkNameField(service, modifications)  {
       var serviceType = modifications.serviceType ? modifications.serviceType : service.serviceType;
       if(isCulvert(serviceType))
-        delete service.nimi;
+        delete service.name;
     }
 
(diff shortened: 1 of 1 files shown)


UI/src/view/point_asset/servicePointForm.js
```diff
@@ -251,14 +251,14 @@
...
+    function checkNameField(service, modifications)  {
       var serviceType = modifications.serviceType ? modifications.serviceType : service.serviceType;
       if(isCulvert(serviceType))
-        delete service.nimi;
+        delete service.name;
     }
 
     function isCulvert(selectedServiceType) {
-      return selectedServiceType.value === 19;
+      return selectedServiceType === 19 || selectedServiceType.value === 19;
```
marquesfm:
```javascript
var serviceType =  selectedServiceType.value ? selectedServiceType.value : selectedServiceType;
return serviceType  === 19;

marquesfm: OR
```javascript;
return (selectedServiceType.value ? selectedServiceType.value : selectedServiceType) === 19;

FilPinto: Applied second option

PR #1502, opened 2020-04-16
DROTH-2375 Value field errors in Kalpa

ricardofmsimoes:
- Fix the Value field in Kalpa for SevenRestrictions


digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/IntegrationApiSpec.scala
```diff
@@ -274,4 +278,27 @@ class IntegrationApiSpec extends FunSuite with ScalatraSuite with BeforeAndAfter
     returntxt should be ("LINESTRING ZM (0.0 0.0 0.0 0.0, 1.0 0.0 0.5 1.0, 4.0 4.0 1.5 6.0)")
 
   }
+
+  test("sevenRestrictionToApi length_limits"){
```
marquesfm, line 282: We can put a better name on this like for example:
"Validate if sevenRestriction JSON generator return all required keys"

marquesfm: Because this sevenRestrictionToApi it's used not only at length_limits

ricardofmsimoes: Applied as suggested

PR #1508, opened 2020-05-06
DROTH-2383 Winter speed limit not working

FilPinto:
-Fixed a bug where possibleValues were not appearing because there was one less argument in the function
-Fixed a bug where WinterSpeedLimit value was being treated as a DynamicValue and not as a NumericValue

### DIFF
--- a/UI/src/view/linear_asset/piecewiseLinearAssetFormElements.js
+++ b/UI/src/view/linear_asset/piecewiseLinearAssetFormElements.js
@@ -6,17 +6,17 @@
     DefaultFormElements: DefaultFormElements
   };
 
-  function DefaultFormElements(unit, editControlLabels, className, authorizationPolicy, defaultValue, possibleValues) {
+  function DefaultFormElements(unit, editControlLabels, className, authorizationPolicy, selectedLinearAsset, defaultValue, possibleValues) {
     var formElem = inputFormElement(unit);
     return formElementFunctions(unit, editControlLabels, className, defaultValue, possibleValues, formElem);
   }
 
-  function WinterSpeedLimitsFormElements(unit, editControlLabels, className, authorizationPolicy, defaultValue, possibleValues) {
+  function WinterSpeedLimitsFormElements(unit, editControlLabels, className, authorizationPolicy, selectedLinearAsset, defaultValue, possibleValues) {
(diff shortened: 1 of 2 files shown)


UI/src/view/linear_asset/piecewiseLinearAssetFormElements.js
```diff
@@ -6,17 +6,17 @@
...
   }
 
-  function WinterSpeedLimitsFormElements(unit, editControlLabels, className, authorizationPolicy, defaultValue, possibleValues) {
+  function WinterSpeedLimitsFormElements(unit, editControlLabels, className, authorizationPolicy, selectedLinearAsset, defaultValue, possibleValues) {
     var formElem = dropDownFormElement(unit);
     return formElementFunctions(unit, editControlLabels, className, defaultValue, possibleValues, formElem);
   }
 
-  function TextualValueFormElements(unit, editControlLabels, className, authorizationPolicy, defaultValue, possibleValues) {
+  function TextualValueFormElements(unit, editControlLabels, className, authorizationPolicy, selectedLinearAsset, defaultValue, possibleValues) {
```
ricardofmsimoes, line 19: why added selectedLinearAsset in the 3 functions if we don't use it inside of them?

FilPinto: You have this in file AssetFormElementsFactory to create the form:
```javascript
return assetFormElementConstructor(asset.layerName)(asset.unit, asset.editControlLabels, asset.className, asset.authorizationPolicy, asset.selectedLinearAsset, asset.defaultValue, asset.possibleValues);
```

In (asset.layerName) you will get the right form to the right asset and the other (.....) will be used by the the specific asset form.
The ones that use this selectedLinearAsset are like this     `function elements(unit, editControlLabels, className, authorizationPolicy, selectedAsset)`

But the others that do not will have bad parameter because defaultValue is the selectedLinearAsset and possibleValues will have the defaultValue so the asset.possibleValues will never appear because the function arguments are max 6 and the parameter pass are 7

ricardofmsimoes: Ok.
Now I understand.

PR #1513, opened 2020-05-13
DROTH-2390 (US 2119) New bus stations, harbours etc. don't show added…

FilPinto:
… by- and last modified-info

-Added added by- and last modified-info to the form

### DIFF
--- a/UI/src/model/selectedMassTransitStop.js
+++ b/UI/src/model/selectedMassTransitStop.js
@@ -100,7 +100,9 @@
         });
     };
 
-    var propertyForMassServiceStops = [
+    var servicePointPropertyOrdering = [
+      'lisatty_jarjestelmaan',
+      'muokattu_viimeksi',
       'palvelu',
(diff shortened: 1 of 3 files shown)


UI/src/model/selectedMassTransitStop.js
```diff
@@ -663,7 +665,8 @@
       isSuggested: isSuggested,
       isAnAddToolOption: isAnAddToolOption,
       isTerminalType: isTerminalType,
-      isServicePointType: isServicePointType
+      isServicePointType: isServicePointType,
+      getServicePointPropertyOrdering: servicePointPropertyOrdering
```
marquesfm, line 669: Keep the same idea already existent on the other methods.
```javascript
getServicePointPropertyOrdering: getServicePointPropertyOrdering

FilPinto: Applied

PR #1514, opened 2020-05-15
Droth 2395 hover doesnt show lane info

FilPinto:


FilPinto:
This implementation is pratically using a general function so one can fetch the lane information of that roadlink(+ filter by the lane code direction if one wishes) and can be used by anyone given lanes(to be used probably on dynamic linear assets with, for example, the road markings asset type).

Tried another type of implementation, one using the already fetched lanes of the getbyboundingbox but the lanePartitioner already needs to filter out the additional lanes or it will give sometimes an error because the geometry is overlapping.

PR #1515, opened 2020-05-20
DROTH-2382 V2

ricardofmsimoes:
- Fix the adjustLanesSideCodes function to remove sub-lanes when sideCode is not ok

### DIFF
--- a/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/lane/LaneService.scala
+++ b/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/lane/LaneService.scala
@@ -238,10 +238,15 @@ trait LaneOperations {
         }
 
         // To remove the lanes with wrong SideCode
-        val toRemove = lanesToProcess.find(lane => (lane.laneCode == MainLane.towardsDirection && lane.sideCode != mainLane11SideCode.value) ||
-                                                    (lane.laneCode == MainLane.againstDirection && lane.sideCode != mainLane21SideCode.value))
-                                      .map(_.id)
-                                      .filterNot(_ == 0L)
+        val toRemove = lanesToProcess.filter { lane =>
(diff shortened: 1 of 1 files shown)


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/lane/LaneService.scala
```diff
@@ -238,10 +238,15 @@ trait LaneOperations {
...
-                                      .filterNot(_ == 0L)
+        val toRemove = lanesToProcess.filter { lane =>
+                                      val isLaneCodeTowards = lane.laneCode >= MainLane.towardsDirection && lane.laneCode <= FourthRightAdditional.towardsDirection
+                                      val isLaneCodeAgainst = lane.laneCode >= MainLane.againstDirection && lane.laneCode <= FourthRightAdditional.againstDirection
+
+                                      (isLaneCodeTowards && lane.sideCode != mainLane11SideCode.value) ||
+                                          (isLaneCodeAgainst && lane.sideCode != mainLane21SideCode.value)
+                                     }
+                                    .map(_.id)
+                                    .filterNot(_ == 0L)
```
RobertVisan, line 249: Shouldn't filter always be done first? For efficiency.

marquesfm: Maybe @RobertVisan is right.
If you don't care about the lanes with id 0 let's filter at the beginning instead of at the end.

Quick suggestion:
```Scala
val toRemove = lanesToProcess.filterNot(_ == 0L).filter { lane =>
```

Or can be put inside the current filter like a first condition to validate.

ricardofmsimoes: Done.

PR #1529, opened 2020-06-03
DROTH-2402 v2

ricardofmsimoes:
- Fix the validation to allow users with municipalities in Ahvenanmaa

### DIFF
--- a/digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/user/User.scala
+++ b/digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/user/User.scala
@@ -1,11 +1,8 @@
 package fi.liikennevirasto.digiroad2.user
 
-import fi.liikennevirasto.digiroad2.Point
-import java.sql.Date
-import java.time.LocalDate
-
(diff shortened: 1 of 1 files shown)


digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/user/User.scala
```diff
@@ -20,6 +17,12 @@ case class Configuration(
                         lastLoginDate: Option[String] = None
                         )
 
+object ElyExceptionsForState {
+  val municipalitiesForAhvenanmaaEly = Set(35, 43, 60, 62, 65, 76, 170, 295, 318, 417, 438, 478, 736, 766, 771, 941)
+
+}
+
+
```
FilPinto, line 25: For me this val can be on the def isAnElyException I don't think there is a need for a object. Another thing is to get this codes from the db but i think you can't access municipalityDao from here

ricardofmsimoes: Yes, we can't access to municipalityDao in this file.
The ideia of the object is with combination of isAnElyException function, we can in the future add more this exceptions in an easy way.

PR #1530, opened 2020-06-04
DROTH-2428 Fixed moving old traffic signs

RobertVisan:


UI/src/view/point_asset/trafficSignForm.js
```diff
@@ -626,7 +627,11 @@
 
     var singleChoiceForPanels = function (property) {
       var publicId = _.head(property);
-      var propertyValue = (!_.isEmpty(_.last(property))) ? '' : _.last(property);
+      var propertyValue;
+      if (_.last(property) === 0)
```
ricardofmsimoes, line 631: Maybe switch to

if ( _.isEmpty( _.last(property) )

UI/src/view/point_asset/trafficSignForm.js
```diff
@@ -626,7 +627,11 @@
...
+        propertyValue = getProperties(me.pointAsset.newAsset.propertyData,  'additional_panel').defaultValue[publicId];
```
ricardofmsimoes, line 632: Can't this getProperties return  undefined ?

UI/src/view/point_asset/trafficSignForm.js
RobertVisan: lodash's isEmpty does not support numbers, so for example isEmpty(1) will return true

UI/src/view/point_asset/trafficSignForm.js
RobertVisan: no, because this property is defined in the assetTypeConfiguration property data for traffic signs

PR #1533, opened 2020-06-09
Droth 2370 ui changes to the lanetool

FilPinto:


UI/src/view/linear_asset/laneModellingStyle.js
```diff
@@ -62,10 +62,21 @@
       return me.lineFeatures(me.getNewFeatureProperties(linearAssets, laneNumber));
     };
 
+    var numberOfLanes = function (asset) {
```
FilPinto, line 65: var numberOfAdditionalLanes seems more recognizable since you are counting only that

UI/src/model/selectedLaneModelling.js
```diff
@@ -4,16 +4,15 @@
...
-    var initial_road_number;
-    var initial_road_part_number;
-    var initial_distance;
-    var end_road_part_number;
-    var end_distance;
+    var roadNumber;
+    var roadPartNumber;
+    var startAddrMValue;
+    var endRoadPartNumber;
+    var endDistance;
```
ricardofmsimoes, line 15: Isnt this endDistance the endAddrMValue ?

UI/src/model/selectedLaneModelling.js
```diff
@@ -334,14 +334,7 @@
...
       var laneIndex = getLaneIndex(laneNumber, marker);
-      var removedLane = self.selection.splice(laneIndex, 1)[0];
-
-      if (removedLane.id !== 0) {
-        //removedLane could be modified by the user so we need to fetch the original
-        var originalRemovedLane = _.find(lanesFetched, {'id': removedLane.id});
-        originalRemovedLane.isDeleted = true;
-        assetsToBeRemoved.push(originalRemovedLane);
-      }
+      self.selection.splice(laneIndex,1);
```
ricardofmsimoes, line 337: Is this ok ? 
Do a slice without saving the result?

FilPinto: We do not need to save the result. Now the delete will work only on lanes that are yet to be created(user is adding and then deleting the lane withou saving it first) so there is no need to send the info to the backend

ricardofmsimoes: So this slice is useless because nothing will be done with the result and the selection will remaining  unchanged.

FilPinto: No, the splice will alter the array it spliced.

ricardofmsimoes: My mistake.

UI/src/model/selectedLaneModelling.js
FilPinto: Yes, are you suggesting to change the name?

UI/src/view/linear_asset/laneModellingStyle.js
FilPinto: Done

UI/src/view/linear_asset/laneModellingForm.js
```diff
@@ -496,12 +542,27 @@
...
+      };
+
       return $('<div class="wrapper read-only">' +
-        '   <div class="form form-horizontal form-dark asset-factory">' +
-        self.userInformationLog() +
-        '   </div>' +
-        '</div>');
+          '   <div class="form form-horizontal form-dark asset-factory">' +
+          '     <div class="form-group">' +
+          '       <p class="form-control-static asset-log-info">Lis&auml;tty j&auml;rjestelm&auml;&auml;n: ' + self.informationLog(info.createdDate, info.createdBy)+ '</p>' +
```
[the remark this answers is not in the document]
FilPinto: copy from dynamic asset form but we can change it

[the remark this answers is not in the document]
FilPinto: Done

UI/src/model/selectedLaneModelling.js
ricardofmsimoes: Ok then.

UI/src/view/linear_asset/laneModellingForm.js
```diff
@@ -554,7 +554,7 @@
       return $('<div class="wrapper read-only">' +
           '   <div class="form form-horizontal form-dark asset-factory">' +
           '     <div class="form-group">' +
-          '       <p class="form-control-static asset-log-info">Lis&auml;tty j&auml;rjestelm&auml;&auml;n: ' + self.informationLog(info.createdDate, info.createdBy)+ '</p>' +
+          '       <p class="form-control-static asset-log-info">Lisätty järjestelmään: ' + self.informationLog(info.createdDate, info.createdBy)+ '</p>' +
```
ricardofmsimoes, line 557: This will print the "Lisätty järjestelmään" correctly? Without any strange characters ?
If you look into dynamicAssetForm.js line 1192, is there this name encoded and make sense to be encoded since this "Nordic"/Special characters aren't standard so we need to encode it for html can parse it correctly.

FilPinto: That's true and I feared that as well when i changed it but works ok, it prints correctly.

marquesfm: I think yes.
We already have cases where we use normally the sentences with tremas without any problems I don't understand why it's used with encoding at the first time...

Like for example:
```Javascript
- var addLeftLane = $('<li>').append($('<button class="btn btn-secondary">Lisää kaista vasemmalle puolelle</button>').click(function() {

- '    <option class="empty" disabled selected>Lisää voimassaoloaika</option>' +

ricardofmsimoes: The enconding is needed or a nice thing to have because in same way you need to have the code repository in encoding windows-1552 to support this tremas.
If you don't encond this characters for HTML you can have issues to show this characters  to some users.
If you want a detail explanation, where one example:
https://text-symbols.com/html/encode/

PR #1558, opened 2020-08-05
Droth 2454 reports from municipality page

FilPinto:


UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
+(function (root) {
+  root.csvReportsWorkList = function () {
```
FilPinto, line 2: Normally we uppercase the first letter, maybe it could negligible...

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -1832,6 +1832,14 @@ class Digiroad2Api(val roadLinkService: RoadLinkService,
     }
   }
 
+  get("/getAssetTypes") {
+    AssetTypeInfo.values.filterNot(_.typeId == 99).toList.sortBy(_.typeId).map { asset =>
+      Map("id" -> asset.typeId,
+        "name" -> { if (asset.nameFI.nonEmpty) asset.nameFI else asset.label }
```
FilPinto, line 1838: Why do you need this "if"? You created this "nameFI" and give the names, is there some asset you did not give the name?

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+        text: "Kaikki kunnat"  /* all municipalities */
+      }));
+
+      _.forEach(municipalities, function (municipality) {
+        $('#municipalities_search').append($('<option>', {
+          value: municipality.id,
+          text: municipality.name
+        }));
+      });
+    }
```
FilPinto, line 68: I think maybe it should be better like this:

```javascript
    function setMunicipalities() {
      var allMunicipalitiesChoice = {id: "1000", name: "Kaikki kunnat"};
      var allMunicipalities = _.concat(municipalities, allMunicipalitiesChoice);

      _.forEach(allMunicipalities , function (municipality) {
        $('#municipalities_search').append($('<option>', {
          value: municipality.id,
          text: municipality.name
        }));
      });
    }
```

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+        }));
+      });
+
+      _.forEach(assetsList, function (asset) {
+        $('#assets_search').append($('<option>', {
+          value: asset.id,
+          text: asset.name
+        }));
+      });
+    }
```
FilPinto, line 90: Same idea as previous suggestion

```javascript
    function setAssets() {
      var topOptions = [
        {id: "1", value: "Kaikki tietolajit"}, /* all assets */
        {id: "2", value: "Pistemäiset tietolajit"}, /* point assets */
        {id: "3", value: "Viivamaiset tietolajit"} /* linear assets */
      ];

      var allAssetOptions = _.concat(topOptions, assetsList);

      _.forEach(allAssetOptions, function (asset) {
        $('#assets_search').append($('<option>', {
          value: asset.id,
          text: asset.name
        }));
      });
    }
```

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+    };
+
+
+
+    function enableSubmitButton() {
+      var isMunicipalityEmpty = _.isEmpty($("#municipalities_search_to, select[name*='municipalityNumbers']").find('option'));
+      var isAssetEmpty = _.isEmpty($("#assets_search_to, select[name*='assetNumbers']").find('option'));
+
+      var disableStatus = !isAssetEmpty && !isMunicipalityEmpty;
+      $('.btn.btn-primary.btn-lg').prop('disabled', !disableStatus);
```
FilPinto, line 242: This looks better
```javascript
var disableStatus = isAssetEmpty || isMunicipalityEmpty;
$('.btn.btn-primary.btn-lg').prop('disabled', disableStatus);
```

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+        4: "images/csv-status-icons/error-icon-small.png",
+        99: "images/csv-status-icons/unknown-error.png"
+      };
+      return '<img src="' + icon[status] + '" title="' + description + '"/>';
+    };
+
+    var scrollbarResize = function () {
+      //if ( $('.job-status tbody tr').length >= 5)
+       // $('.job-status thead').css("width", "calc(100% - 17px)");
+    };
```
FilPinto, line 486: Commented code

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+
+      if(!_.isEmpty(jobsInProgress)) {
+        backend.getExportsJobsByIds(jobsInProgress.toArray()).then(function(jobs){
+          var endedJobs = _.filter(jobs, function(job){return job.status !== 1;});
+          _.map(endedJobs, replaceRow);
+        });
+      } else {
+        clearInterval(refresh);
+        refresh = null;
+      }
```
FilPinto, line 312: There is no need to negate the _.isEmpty
```javascript
     if(_.isEmpty(jobsInProgress)) {
        clearInterval(refresh);
        refresh = null;
      } else {
        backend.getExportsJobsByIds(jobsInProgress.toArray()).then(function(jobs){
          var endedJobs = _.filter(jobs, function(job){return job.status !== 1;});
          _.map(endedJobs, replaceRow);
        });
      }
```

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  get("/logs/:id") {
+    val id = params("id").toLong
+    csvDataExporter.getById(id).getOrElse("Logia ei löytynyt.")
+  }
+
+  get("/logs/:ids") {
+    val ids = params("ids").split(',').map(_.toLong).toSet
+    csvDataExporter.getByIds(ids)
+  }
```
FilPinto, line 72: They are same entrypoints no?

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/CsvDataExporter.scala
```diff
@@ -0,0 +1,108 @@
...
+    def update(id: Long, status: Status, content: Option[String] = None, withTransaction: Boolean = true) : Long  = {
+      if (withTransaction)
+        withDynTransaction {
+          exportReportDAO.update(id, status, content)
+        }
+      else
+        exportReportDAO.update(id, status, content)
+    }
+
+    def createStandardCSVData(headers: Seq[String], values: Seq[Map[String, String]]): String = {
```
FilPinto, line 75: Is this function used only in the Spec? Then we should move it to that in my opinion

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/CsvDataExporter.scala
```diff
@@ -0,0 +1,108 @@
...
+
+      if (headers.isEmpty && values.isEmpty)
+        return ""
+
+      val headerLine = if(headers.nonEmpty) headers.toList.mkString(";").concat("\r\n")
+                        else ""
+
+      val rows = if (values.isEmpty) {
```
FilPinto, line 83: This if is unnecessary since the .map on a empty list will return a empty list

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/CsvDataExporter.scala
```diff
@@ -0,0 +1,108 @@
...
+                }
+                else{
+                  values.map { mapValues =>
+                    if (headers.isEmpty) {
+                      mapValues.values.toList
+                    }
+                    else {
+                      var elems = Seq[String]()
+                      headers.foreach{h => elems = elems ++ Seq(mapValues(h)) }
+                      elems
```
FilPinto, line 94: You can use a fold left here
```scala
headers.foldLeft(Seq[String]()) { case (elems, h) =>
    elems ++ Seq(mapValues(h))
}
```

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/AssetReportCsvExporterSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+      val generatedDataId = assetReportCsvExporter.exportAssetsByMunicipalityCSVGenerator(assetTypes, municipalities, logId, false)
+
+      val result = csvDataExporter.getById(generatedDataId, false)
+      val finalContent = result match{
+        case Some(x) => x.content match {
+          case Some(content) => content
+          case _ => ""
+        }
+        case _ => ""
+      }
```
FilPinto, line 93: For me, in a test, this is useless. You are testing tha it should contains something in there so no need to alter it to a empty string because of options. If is None for some reason the test will fail because of a exception and that's fine because it should have a Some().
 
var finalContent = result.get.content.get

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/AssetReportCsvExporterSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+      val generatedDataId = assetReportCsvExporter.exportAssetsByMunicipalityCSVGenerator(assetTypes, municipalities, logId, false)
+
+      val result = csvDataExporter.getById(generatedDataId, false)
+      val finalContent = result match{
+        case Some(x) => x.content match {
+                        case Some(content) => content
+                        case _ => ""
+                      }
+        case _ => ""
+      }
```
FilPinto, line 147: Same idea as previous suggestion

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/CsvDataExporterSpec.scala
```diff
@@ -0,0 +1,69 @@
+package fi.liikennevirasto.digiroad2.dataexport
```
FilPinto, line 1: I don't know why you are testing this createStandardCSVData. There is no other place that method is being used except here

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+
+      (pointAssets, linearAssets) match {
+        case (true, false) =>
+          (allPoints ++ assetTypesList).filterNot(_ == PointAssets.id).toList
+
+        case (false, true) =>
+          (allLinears ++ assetTypesList).filterNot(_ == LinearAssets.id).toList
+
+        case (true, true) =>
+          (allPoints ++ allLinears).filterNot(id => id == PointAssets.id || id == LinearAssets.id).toList
```
FilPinto, line 215: (allPoints ++ allLinears) should be enough since you are not adding the assetTypesList so the ids 1/2 will not appear

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+  def decodeAssetsToProcess(assetTypesList: List[Int]): List[Int] = {
+
+    val ( pointAssets, linearAssets ) = ( assetTypesList.contains(PointAssets.id), assetTypesList.contains(LinearAssets.id) )
+
+    if ( assetTypesList.contains(AllAssets.id) ) {
+      AssetTypeInfo.values.map(_.typeId).toList
+    }
+    else if ( pointAssets || linearAssets ) {
+
+      val (linearAssetType, pointAssetType) = AssetTypeInfo.values.partition( _.geometryType == "linear" )
```
FilPinto, line 202: AssetTypeInfo.values.toList.partition(...) so there is no need to put List in this else if all over the place

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+      result.toList
+    }
+
+    output.mkString("\r\n")
+  }
+
+
+  def decodeAssetsToProcess(assetTypesList: List[Int]): List[Int] = {
```
FilPinto, line 193: I would prefer to return a Set[Int] so no asset type Id is equal, just in case. But this is fine too

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+      }
+
+    }
+    else
+      assetTypesList
+  }
+
+
+  def decodeMunicipalitiesToProcess(municipalities: List[Int]): List[Int] = {
+    if (municipalities.isEmpty || municipalities.contains(1000)) { /* 1000 = All municipalities */
```
FilPinto, line 228: Is possible to send a empty list of municipalities?

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+    try {
+
+      val finalAssetList = decodeAssetsToProcess(assetTypesList)
+      val (linearAssets, pointAssets) = finalAssetList.partition( AssetTypeInfo(_).geometryType == "linear")
+
+      val linearContent = if (linearAssets.nonEmpty) exportAssetsByMunicipality(assetTypesList, municipalitiesList, withTransaction = withTransaction)
+                        else List()
+
+      val pointContent = if (pointAssets.nonEmpty) exportAssetsByMunicipality(assetTypesList, municipalitiesList, extractPointsAsset = true, withTransaction)
+                        else List()
```
FilPinto, line 249: Pattern matching could be what you are looking here

val (linearContent,  pointContent) = (linearAssets.nonEmpty, pointAssets.nonEmpty) match ...

or even val allContent = (same match above) since you are not looking to do anything with the linear/point content

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+        }
+      }
+      else{
+        val municipalityInfo = municipalityDao.getMunicipalityIdByName(municipalityName)
+        assetReporterDAO.getTotalTrafficSignNewLaw(municipalityInfo.head.id)
+      }
+    }
+
+
+    val output = if (exportAssetReports.isEmpty) {
```
FilPinto, line 157: I don't think this if isEmpty is necessary. The .map will return a empty list in it's own

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+
+          (userResults, municResults)
+        }
+      }
+      else{
+        val userResults = userProvider.getUsers()
+        val municResults = municipalityDao.getMunicipalitiesNameAndIdByCode(municipalitiesList.toSet)
+
+        (userResults, municResults)
+      }
```
FilPinto, line 89: Move this code to a function since is repeated
    
```scala
 val userResults = userProvider.getUsers()
        val municResults = municipalityDao.getMunicipalitiesNameAndIdByCode(municipalitiesList.toSet)

        (userResults, municResults)
```

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -1832,6 +1832,14 @@ class Digiroad2Api(val roadLinkService: RoadLinkService,
     }
   }
 
+  get("/getAssetTypes") {
+    AssetTypeInfo.values.filterNot(_.typeId == 99).toList.sortBy(_.typeId).map { asset =>
```
RobertVisan, line 1836: instead of '99' you have on asset.scala an UnknownAssetTypeId that extends AssetTypeInfo with typeId = 99

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
RobertVisan: In case you you want to maintain this implementation 
``` Scala
 "name" -> { if (asset.nameFI.nonEmpty) asset.nameFI else asset.label }
```
You might want to trim before you check for nonEmpty

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
RobertVisan, line 72: This here won't work. They are using the same entry point so it will always match with the same one. Since this is using ScalatraBase it will order the entry points from the end of the document to the start therefore it will always match with 
``` Scala 
get("/logs/:ids") {

File: digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/AssetReportCsvExporterSpec.scala [hunk 23]
```diff
@@ -0,0 +1,160 @@
...
+
+  val operatorUser = User(26867454, "k647320", Configuration(None, None, None, None, None, Set(), Set(), Set("operator"), Some("2018-11-27"), None), None)
+
+  val users = List( operatorUser,
+                  User(642513, "l485693", Configuration(None, None, None, None, None, Set(408, 739), Set(), Set(), None, None), None),
+                  User(123456, "lx775599", Configuration(None, None, None, None, None, Set(), Set(), Set("elyMaintainer"), Some("2019-12-07"), None), None) )
+
+
+  test("test function exportAssetsByMunicipality") {
+    OracleDatabase.withDynTransaction {
```
RobertVisan, line 60: Instead of doing this and at the end using ``` dynamicSession.rollback()``` it looks better with
``` Scala
private def runWithRollback(test: => Unit): Unit = TestTransactions.runWithRollback()(test)
```
And use ```runWithRollback``` just like we do in the other tests

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
RobertVisan: Just a reminder that we shouldn't use pattern matching to only check for true or false because it is heavier than an if else

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/asset/asset.scala
```diff
@@ -913,51 +915,52 @@ object AssetTypeInfo {
...
+case object TrHeightLimit extends  AssetTypeInfo {val typeId = 360; def geometryType = "point"; val label = "TrHeightLimit"; val layerName = "trHeightLimits"; val nameFI = "Tierekisteri Suurin sallittu korkeus"}
+case object TrWidthLimit extends  AssetTypeInfo {val typeId = 370; def geometryType = "point"; val label = "TrWidthLimit"; val layerName = "trWidthLimits"; val nameFI = "Tierekisteri Suurin sallittu leveys"}
+case object Manoeuvres extends AssetTypeInfo { val typeId = 380; def geometryType = "linear"; val label = "Manoeuvre"; val layerName = "manoeuvre"; val nameFI = "Kääntymisrajoitus" }
+case object CareClass extends  AssetTypeInfo {val typeId = 390; def geometryType = "linear"; val label = "CareClass"; val layerName = "careClass"; val nameFI = "Hoitoluokat"}
+case object CarryingCapacity extends AssetTypeInfo { val typeId = 400; def geometryType = "linear"; val label = "CarryingCapacity" ; val layerName = "carryingCapacity"; val nameFI = "Kantavuus"}
+case object AnimalWarnings extends AssetTypeInfo { val typeId = 410; def geometryType = "linear"; val label = "AnimalWarnings" ; val layerName = "animalWarnings"; val nameFI: String = "Elainvaroitukset"}
```
FilPinto, line 958: val nameFI: String is not necessary you can just put val nameFI

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/ExportReportDAO.scala
```diff
@@ -0,0 +1,86 @@
...
+                                      .map( asset => AssetTypeInfo(asset.toInt).nameFI )
+                                      .mkString(",")
+
+      val municipalitiesAsSet = municipalities.split(",").map(_.toInt).toSet
+      val municipalitiesName = municipalityDao.getMunicipalitiesNameAndIdByCode( municipalitiesAsSet )
+                                              .map(_.name)
+                                              .mkString(",")
+
+
+      ExportStatusInfo(id, status, Status.apply(status).descriptionFi, fileName, createdBy, createdDate, assetsName, municipalitiesName, content)
```
RobertVisan, line 38: Minor thing but I don't think is necessary to specifically call apply, just ``` Status(status).descriptionFi ``` should work the same

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
+(function (root) {
+  root.csvReportsWorkList = function () {
+    WorkListView.call(this);
+    var me = this;
+    this.hrefDir = "#work-list/csvReports";
```
RobertVisan, line 5: Can you check if this is being used

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+    this.title = 'Raportointityökalu';
+    var backend;
+    var municipalities;
+    var assetsList;
+    var refresh;
+
```
RobertVisan, line 11: We can build a structure here, something immutable, to have an easy access to the id and value from assets like this, feel free to change the names
``` JavaScript
var options = {
     assets: {id: "1", value: "Kaikki tietolajit"}, 
     points: {id: "2", value: "Pistemäiset tietolajit"}, 
     linears: {id: "3", value: "Viivamaiset tietolajit"}
}
```
Here we also have the option to add ```municipalities: {id: "1000", value: "Kaikki kunnat"} ``` and distinguish it with a boolean value in the same structure or just build it separate from the others. 
This structure can now be accessed like this 
``` JavaScript
options["points"].id
```
We can use this structure in the code when needed. I will be pointing where this can be used in the comments

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+          });
+        } else
+          setAssets();
+      }
+    }
+
+    function setMunicipalities() {
+      $('#municipalities_search').append($('<option>', {
+        value: "1000",
+        text: "Kaikki kunnat"  /* all municipalities */
```
RobertVisan, line 59: We can use the structure here

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+        {"id": "1", "value": "Kaikki tietolajit"}, /* all assets */
+        {"id": "2", "value": "Pistemäiset tietolajit"}, /* point assets */
+        {"id": "3", "value": "Viivamaiset tietolajit"} /* linear assets */
+      ];
+
+      _.forEach(topOptions, function (option) {
+        $('#assets_search').append($('<option>', {
+          value: option.id,
+          text: option.value
+        }));
```
RobertVisan, line 81: We can use the structure here

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+    }
+
+
+    function populatedMultiSelectBoxes() {
+      getMunicipalities();
+      getAssetsType();
+    }
+
+    function municipalitySort(a, b){
+      /* value: "1000", text: "Kaikki kunnat"  /* all municipalities */
```
RobertVisan, line 99: We can use the structure here

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+
+      return a.text > b.text ? 1 : -1;
+    }
+
+    function assetTypeSort(a, b){
+      /* value: "1", text: "Kaikki tietolajit"}, /* all assets */
+      /* value: "2", text: "Pistemäiset tietolajit"}, /* point assets */
+      /* value: "3", text: "Viivamaiset tietolajit"} /* linear assets */
+
+      var topElems = ["1", "2", "3"];
```
RobertVisan, line 114: We can use the structure here

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: My mistake.

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
ricardofmsimoes: > Why do you need this "if"? You created this "nameFI" and give the names, is there some asset you did not give the name?

But for some reason if no one fill that value or it is empty, I will use the label as a backup attempt to send an empty description to frontend.

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: don't understand the advantage on that approach or leave as is

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/CsvDataExporter.scala
ricardofmsimoes: Yes.
This function is to create a standard CSV export.
We don't have for now a export that is a standard csv, but the code is here to be ready in future.
However it maybe require some adjustments

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/CsvDataExporter.scala
ricardofmsimoes: If I can easly check if the list is empty or not why should leave that to be validate in .map ?

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/AssetReportCsvExporterSpec.scala
ricardofmsimoes: what is preferable? A failed test due a should condition that is false or a test fail due a None.get exception?

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/AssetReportCsvExporterSpec.scala
ricardofmsimoes: Same reply as previous

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/CsvDataExporterSpec.scala
ricardofmsimoes: It is a "generic" class for standard csv creation.
So I think it the test make all sense.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
ricardofmsimoes: well...it is bug :D

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
ricardofmsimoes: > Pattern matching could be what you are looking here
> 
> val (linearContent, pointContent) = (linearAssets.nonEmpty, pointAssets.nonEmpty) match ...
> 
> or even val allContent = (same match above) since you are not looking to do anything with the linear/point content

Don't know the vantage on that.
Because I will create 4 cases inside a match when 2 IF...else do the job in a simple way.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
ricardofmsimoes: If I can easly check if the list is empty or not why should leave that to be validate in .map ?

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/asset/asset.scala
ricardofmsimoes: was left by mistake.

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: I think it is by WorkList

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/CsvDataExporter.scala
FilPinto: That's true, I remember now we discussed that, my bad.

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/CsvDataExporterSpec.scala
FilPinto: That's true, I remember now we discussed that, my bad.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
FilPinto: Yeah you are right I was thinking too much Scala-like in this one.

UI/src/application.js
```diff
@@ -138,6 +138,7 @@
     new SpeedLimitWorkList().initialize();
     new InaccurateWorkList().initialize();
     new PrivateRoadsWorkList().initialize(backend);
+    new csvReportsWorkList().initialize(backend);
```
marquesfm, line 141:
```JavaScript
new CsvReportsWorkList().initialize(backend);

File: digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala (hunk 15 above)
Comment (marquesfm) [reply digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala:215]: I disagree with you @FilPinto.
If we reach this point it's because the user sent 2 and 3 in assetTypesList List, if that is possible we will enter here and we will get the points and the linears and we need to exclude those dummy values from the list.

What we can do better is ignore the Case(true, true) and do something like this:
```Scala
      (pointAssets, linearAssets) match {
        case (true, false) =>
          (allPoints ++ assetTypesList).filterNot(_ == PointAssets.id).toList

        case (false, true) =>
          (allLinears ++ assetTypesList).filterNot(_ == LinearAssets.id).toList

        case _ =>
          AssetTypeInfo.values.map(_.typeId).toList
      }

File: digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala (hunk 18 above)
Comment (marquesfm) [reply digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala:228]: Validate if the Municipalities list is not empty when receive it and then:

```Scala
  def decodeMunicipalitiesToProcess(municipalities: List[Int]): List[Int] = {
    if (municipalities.contains(1000)) { /* 1000 = All municipalities */
      withDynSession {
        municipalityDao.getMunicipalitiesInfo.map(_._1).toList
      }
    }
    else
      municipalities
  }

File: digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala (hunk 2 above)
Comment (FilPinto) [reply digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala:1838]: Ok

File: digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala [hunk 33]
```diff
@@ -0,0 +1,264 @@
...
+            Some( ExportAssetReport(municipalitiesInfo.filter(_.id == municipality).head.name, firstElem.assetType, currentAssetType.size,
+                              firstElem.assetNameFI, firstElem.assetGeometryType, operatorUser, operatorDate, municipalityUser, municipalityDate) )
+          }
+          else
+            None
+        }
+
+      }
+      else
+        None
```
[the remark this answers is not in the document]
ricardofmsimoes: Maybe we can use some ideas of the suggestion but not as is.
Because if a municipality don't have an asset in the list of the extraction we don't want to send the information of that asset with value 0.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+    }
+  }
+
+
+  def extractRecords( exportAssetReports: List[ExportAssetReport], withTransaction: Boolean = true): String = {
+
+    def getTotalNewLawTrafficSigns(municipalityName: String): Int = {
+      if(withTransaction) {
+        withDynSession {
+          val municipalityInfo = municipalityDao.getMunicipalityIdByName(municipalityName)
```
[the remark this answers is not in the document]
ricardofmsimoes: This is an old method.
And it is being called when the users request the traffic sign information (and not to all assets)

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/middleware/DataExportManager.scala
```diff
@@ -0,0 +1,21 @@
...
+case class CsvDataExporterInfo(assetTypeList: List[Int], municipalitiesList: List[Int], fileName: String, user: User,  logId: Long)
+
+
+class DataExportManager(roadLinkService: RoadLinkService, eventBus: DigiroadEventBus, userProvider: UserProvider) {
+
+  lazy val assetReportCsvExporter: AssetReportCsvExporter = new AssetReportCsvExporter(roadLinkService, eventBus, userProvider)
+
+
+  def processExport(data: CsvDataExporterInfo): Unit ={
+    assetReportCsvExporter.exportAssetsByMunicipalityCSVGenerator(data.assetTypeList,data.municipalitiesList, data.logId);
```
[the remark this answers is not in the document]
ricardofmsimoes: Done

UI/src/application.js
ricardofmsimoes: Done

UI/test/integration-tests.html
```diff
@@ -220,6 +220,7 @@
 <script type="text/javascript" src="../src/view/speedLimitWorkList.js"></script>
 <script type="text/javascript" src="../src/view/inaccurateWorkList.js"></script>
 <script type="text/javascript" src="../src/view/privateRoadsWorkList.js"></script>
+<script type="text/javascript" src="../src/view/csvReportsWorkList.js"></script>
```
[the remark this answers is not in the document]
ricardofmsimoes: Added.

UI/src/less/core/forms.less
```diff
@@ -698,6 +698,72 @@ input[type="checkbox"] {
...
+
+    .labelBoxRight {
+      width: 300px !important;
+      text-align: center;
+    }
+
+    .labelBoxLeft {
+      width: 300px !important;
+      text-align: center;
+    }
```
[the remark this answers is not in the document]
ricardofmsimoes: Done

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: Fixed

UI/src/view/csvReportsWorkList.js
```diff
@@ -0,0 +1,489 @@
...
+        });
+
+        $(".job-status-table").find("#" + job.id + ".btn-download").on('click', function (event) {
+          downloadCsv(event);
+        });
+      }
+    }
+
+    var hideImporter = function() {
+      $('#csvExport').hide();
```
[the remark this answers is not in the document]
ricardofmsimoes: No. showImport is the function who got the show wrong.

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: function removed.

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
ricardofmsimoes: Done.

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
ricardofmsimoes: Added the trim

digiroad2-geo/src/main/scala/fi/liikennevirasto/digiroad2/asset/asset.scala
ricardofmsimoes: Fixed

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/ExportReportDAO.scala
ricardofmsimoes: Applied

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: applied.

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: apllied.

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: applied

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: applied

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: done

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: applied

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: applied

UI/src/view/csvReportsWorkList.js
ricardofmsimoes: done

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
```diff
@@ -0,0 +1,101 @@
...
+    } else {
+      val data = csvDataExporter.getExportById(id.toLong)
+
+      data match {
+        case  Some(info) => info
+        case _ =>
+      }
+    }
+
+  }
```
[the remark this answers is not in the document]
ricardofmsimoes: Done

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  get("/log") {
```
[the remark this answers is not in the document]
ricardofmsimoes: Done.

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
```diff
@@ -0,0 +1,101 @@
...
+    csvDataExporter.getByUser(userProvider.getCurrentUser().username)
+  }
+
+  get("/logs/:id") {
+    val id = params("id").toLong
+    csvDataExporter.getById(id).getOrElse("Logia ei löytynyt.")
+  }
```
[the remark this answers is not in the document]
ricardofmsimoes: removed.

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
```diff
@@ -0,0 +1,101 @@
...
+    eventBus.publish("exportCSVData", CsvDataExporterInfo(assetTypes, municipalities, filename, user, logId) )
+
+    csvDataExporter.getById(logId)
+  }
+
+  def validateOperation(): Unit = {
+    if(!(userProvider.getCurrentUser().isOperator())) {
+      halt(Forbidden("Vain operaattori voi suorittaa tämän toiminnon"))
+    }
+  }
```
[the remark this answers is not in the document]
ricardofmsimoes: Done.

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
```diff
@@ -0,0 +1,101 @@
...
+    csvDataExporter.getByIds(ids)
+  }
+
+  post("/generateCsvReport/:municipalities/:assetTypes") {
+    val municipalitiesParam = params("municipalities")
+    val assetTypesParam = params("assetTypes")
+
+    validateOperation()
+
+    val user = userProvider.getCurrentUser()
```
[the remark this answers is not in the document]
ricardofmsimoes: Done.

UI/src/utils/backend-utils.js
```diff
@@ -531,6 +531,37 @@
...
+      return $.getJSON('api/getAssetTypes');
+    };
+
+    this.getCsvReport = function(id){
+      return $.getJSON('api/export/downloadCsv/' + id);
+    };
+
+    this.getExportJobById = function(id) {
+      return $.getJSON('api/export/log/' + id);
+    };
```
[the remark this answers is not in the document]
ricardofmsimoes: Removed

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
ricardofmsimoes: Done

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
ricardofmsimoes: apllied

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
ricardofmsimoes: done

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
ricardofmsimoes: as requested

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -0,0 +1,264 @@
...
+    val operatorUsers = users.filter(_.isOperator())
+    val operatorUsersUsername = operatorUsers.map(_.username)
+
+    municipalitiesList.flatMap{ municipality =>
+
+      val municipalityUsers = users.filter( _.isAuthorizedToWrite(municipality) )
+                                    .filterNot( user => operatorUsersUsername.contains(user.username) )
+                                    .map( _.username )
+
+      val linearAssets = extractData(municipality)
```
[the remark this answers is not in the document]
ricardofmsimoes: Done.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
ricardofmsimoes: Function created to avoid repeated code

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/CsvDataExporter.scala
ricardofmsimoes: Apllied.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/CsvDataExporter.scala
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    def getExportById(id: Long, withTransaction: Boolean = true) : Option[ExportStatusInfo]  = {
+      if (withTransaction)
+        withDynTransaction {
+          exportReportDAO.get(id)
+        }
+      else
+        exportReportDAO.get(id)
+    }
```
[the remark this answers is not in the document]
ricardofmsimoes: getExportById removed

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/AssetReportCsvExporterSpec.scala
ricardofmsimoes: Done.

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/AssetReportCsvExporterSpec.scala
```diff
@@ -0,0 +1,160 @@
...
+      val municipalitiesParam = "408, 739"
+      val assetTypesParam = "130"
```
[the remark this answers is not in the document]
ricardofmsimoes: Done.

digiroad2-api-oth/src/test/scala/fi/liikennevirasto/digiroad2/dataexport/AssetReportCsvExporterSpec.scala
ricardofmsimoes: done

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
ricardofmsimoes: fixed.

ricardofmsimoes: fixed.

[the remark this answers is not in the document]
ricardofmsimoes: I don't agree.
Because it is only used to get the extractions that are "in-progress" .
For user I think it will not be more than 1 or 2 Ids will be sent in URL

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  get("/logs/:ids") {
+    val ids = params("ids").split(',').map(_.toLong).toSet
+    csvDataExporter.getByIds(ids)
+  }
+
+  post("/generateCsvReport/:municipalities/:assetTypes") {
+    val municipalitiesParam = params("municipalities")
+    val assetTypesParam = params("assetTypes")
```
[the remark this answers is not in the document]
ricardofmsimoes: Done.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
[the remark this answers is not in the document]
ricardofmsimoes: As previously said, partial of suggestion implemented.
The reason for that is not to send rows with value zero because in the user story the example seems in that way to not send zeros

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/ExportReportDAO.scala
```diff
@@ -0,0 +1,95 @@
...
+                        }
+
+      val municipalitiesName = if ( municipalities == null ) //can occur with status = 2
+                                  ""
+                                else {
+                                  val municipalitiesAsset = municipalities.split(",").map(_.toInt).toSet
+                                  municipalityDao.getMunicipalitiesNameAndIdByCode(municipalitiesAsset)
+                                                .map(_.name)
+                                                .mkString(",")
+                                }
```
FilPinto, line 44: If it can be null then you can us r.nextStringOption() in both exportedAssets and municipalities .

You then can do the .getOrElse("").split........

digiroad2-oracle/src/main/resources/db/migration/V232__add_export_report_table.sql
```diff
@@ -0,0 +1,20 @@
...
+  CONSTRAINT export_report_pk PRIMARY KEY (ID)
+);
+
+CREATE INDEX export_report_created_by_idx ON export_report (created_by);
+
+/* Obstacles Asset = 220
+   Change the name from esterakennelma to Esterakennelma */
+UPDATE ASSET_TYPE
+SET name = TRIM(INITCAP(name))
+WHERE ID = 220;
```
FilPinto, line 20: Check if another part of the code this needs to change too, for example in ObstacleService.scala there is a typePublicId = "esterakennelma" that probably needs to be changed too.

UI/src/view/csvReportsWorkList.js
FilPinto: WorkListView does not have this variable at all and in this csvReportsWorkList this is not being used so I will delete it.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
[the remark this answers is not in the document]
FilPinto: This method returns a list because a municipality name is not specific unique. In the moment it is, but there is no rule that municipalities shouldn't have equal names.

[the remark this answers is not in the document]
FilPinto: Changed the getMunicipalityIdByName to another name and returns now a option

digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/dataexport/ExportDataApi.scala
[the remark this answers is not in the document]
FilPinto: Changed ids to be a query paramenter and not a variable in the path

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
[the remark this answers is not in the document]
FilPinto: Refactored the code

PR #1564, opened 2020-08-18
DROTH-2508 Restore expired assets in Ahvenanmaa

RobertVisan:
-Add Ahvenanmaa municipalities as exceptions to TR-import batches

### DIFF
--- a/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/OracleAssetDao.scala
+++ b/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/OracleAssetDao.scala
@@ -127,4 +127,24 @@ class OracleAssetDao {
     Queries.updateAssetModified(id, username).first
     sqlu"update asset set valid_to = sysdate where id = $id".first
   }
+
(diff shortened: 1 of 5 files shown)


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/OracleAssetDao.scala
```diff
@@ -127,4 +127,24 @@ class OracleAssetDao {
...
+            where a.asset_type_id in (#${assetTypeIdsToRestore.mkString(",")}) and (a.valid_to between to_date($startDate, 'yyyy-mm-dd hh24:mi:ss') and to_date($endDate, 'yyyy-mm-dd hh24:mi:ss'))
+         """.as[Long].list
+    }
+  }
+
+  def restoreExpiredAssetsByIds(assetsToBeRestored: Set[Long], username: String) = {
+    sqlu"""
+          update asset
+          set modified_date = sysdate, modified_by = $username, valid_to = null
+          where id in (#${assetsToBeRestored.mkString(",")})""".execute
```
FilPinto, line 148: Can you guarantee that the assetsToBeRestored is maximum of 1000? Because if not it will fail since the in operand can only contain 1000 cases.

RobertVisan: Applied

PR #1573, opened 2020-08-28
DROTH-2521 Report lists changes in roads which it should not

FilPinto:
-Add filter from assets on municipality road links
-Fixed missing lanes and manoeuvres on report
-Refactor code

### DIFF
--- a/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
+++ b/digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
@@ -55,18 +55,12 @@ class AssetReportCsvExporter(roadLinkServiceImpl: RoadLinkService, eventBusImpl:
         .map(_.linkId)
 
       if (extractPointsAsset) {
-        if(assetTypesList.size == 1 && assetTypesList.head == ServicePoints.typeId) {
-          assetReporterDAO.servicePointQuery(Seq(municipality))
-        } else {
-          val pointAssetReports = assetReporterDAO.pointAssetQuery(linkIds, assetTypesList.filterNot(_ == ServicePoints.typeId))
-
+          val pointAssetReports = assetReporterDAO.pointAssetQuery(linkIds, assetTypesList)
           val servicePointAssetReports = if(assetTypesList.contains(ServicePoints.typeId)) assetReporterDAO.servicePointQuery(Seq(municipality)) else List()
 
           pointAssetReports ++ servicePointAssetReports
-        }
       } else {
         val linearAssetReports = assetReporterDAO.linearAssetQuery(linkIds, assetTypesList)
-
         val laneAssetReports = if(assetTypesList.contains(Lanes.typeId)) assetReporterDAO.laneQuery(linkIds) else List()
         val manoeuvreAssetReports = if(assetTypesList.contains(Manoeuvres.typeId)) assetReporterDAO.manoeuvreQuery(linkIds) else List()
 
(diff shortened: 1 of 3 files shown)


digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -50,14 +50,27 @@ class AssetReportCsvExporter(roadLinkServiceImpl: RoadLinkService, eventBusImpl:
...
-      }
-      else {
-        val linkIds = roadLinkServiceImpl.getVVHRoadLinksF(municipality)
-                                         .map(_.linkId)
+        if(assetTypesList.size == 1 && assetTypesList.head == ServicePoints.typeId) {
+          assetReporterDAO.servicePointQuery(Seq(municipality))
+        } else {
+          val pointAssetReports = assetReporterDAO.pointAssetQuery(linkIds, assetTypesList.filterNot(_ == ServicePoints.typeId))
+
+          val servicePointAssetReports = if(assetTypesList.contains(ServicePoints.typeId)) assetReporterDAO.servicePointQuery(Seq(municipality)) else List()
```
RobertVisan, line 63: We can do a partition here on assetTypeList to divide servicePoints and pointAssets to avoid going through the assetTypeList two times on filter and contains. It will remain something like this 
``` Scala
val (servicePointsList, pointAssetList) = assetTypesList.partition(_ == ServicePoints.typeId)

val pointAssetReports = assetReporterDAO.pointAssetQuery(linkIds, pointAssetList)

val servicePointAssetReports = if(servicePointsList.nonEmpty) assetReporterDAO.servicePointQuery(Seq(municipality)) else List()
```

RobertVisan (approved):

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
```diff
@@ -50,14 +50,27 @@ class AssetReportCsvExporter(roadLinkServiceImpl: RoadLinkService, eventBusImpl:
...
+
+          pointAssetReports ++ servicePointAssetReports
+        }
```
marquesfm, line 66: At this point, we don't need to perform the second IF clause if we modify the query inside of the pointAssetQuery method, and the "filterNot(_ == ServicePoints.typeId)" can be deleted too.

Because, since the service points don't have any relation with LRM_POSITION, the query inside of the pointAssetQuery method will return nothing when the ServicePoints type id is selected.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/csvexporter/AssetReporterDAO.scala
```diff
@@ -41,23 +41,58 @@ class AssetReporterDAO {
...
+        JOIN ASSET_LINK al ON a.id = al.asset_id
+        JOIN LRM_POSITION pos ON al.position_id = pos.id
+        JOIN #$idTableName i ON i.id = pos.link_id
       WHERE (a.valid_to IS NULL OR a.valid_to > SYSDATE)
       AND at2.GEOMETRY_TYPE = 'point'
       #$filter
       """.as[AssetReport](getResult).list
     }
-
   }
```
marquesfm, line 83: In this method, we need to remove the if where it's checked if assetTypesList is empty or not because they are mandatory, at least one asset type.
We can put the ELSE of that verification directly at the code without any problem.

marquesfm (commented):
OK with some suggestions.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
FilPinto: Applied another suggestion that modifies the behaviour you mentioned

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/dao/csvexporter/AssetReporterDAO.scala
FilPinto: Applied

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/csvDataExporter/AssetReportCsvExporter.scala
FilPinto: Applied

PR #1574, opened 2020-08-28
Droth-2523 Visual upgrade to report table

RobertVisan:


UI/src/less/core/forms.less
```diff
@@ -634,6 +634,10 @@ input[type="checkbox"] {
         display: none;
       }
     }
+    .multiSelectItems{
+      text-align: center;
+      margin-left: -10px
```
FilPinto, line 639: Maybe doesn't cause anything but add the ';' in this

margin-left: -10px;

marquesfm (approved):

UI/src/less/site/pages.less
```diff
@@ -129,3 +145,13 @@
...
+
+.bigListPopUp {
+  overflow-y: auto;
+  height: auto;
+  max-height: 500px;
+}
+
+.clickableText {
+  cursor: pointer;
+}
```
FilPinto, line 157: You already have this one you can use:

.selectable{
  cursor: pointer;
}

UI/src/less/core/forms.less
RobertVisan: Done

UI/src/less/site/pages.less
RobertVisan: clickableText removed and selectable applied

PR #1575, opened 2020-08-28
DROTH-2524 Additional signs with time range

marquesfm:
- Fix text position of the valid period additional signs;
- Create new validation for valid period additional signs form.


UI/src/assetTypeConfiguration.js
```diff
@@ -1393,10 +1411,32 @@
...
+          if (_.isEmpty(additionalPanelsProps.values)) {
+            isValidAdditionalSign = true;
+          } else {
+            isValidAdditionalSign = _.every(additionalPanelsProps.values, function (additionalPanelProp) {
+              var additionalSignsValidations = _.find(validations, function (validation) {
+                return _.includes(validation.types, additionalPanelProp.panelType);
+              });
+
+              var additionalPanelValue = additionalPanelProp.panelValue;
+              var isValidPanelValue = additionalSignsValidations ? additionalSignsValidations.validate(additionalPanelValue, additionalSignsValidations.regexToUse) : true;
```
RobertVisan, line 1431: We can return the value here.

RobertVisan (approved):

UI/src/assetTypeConfiguration.js
```diff
@@ -1393,10 +1411,32 @@
...
+
+              return isValidPanelValue;
+            });
+          }
```
FilPinto, line 1435: I don't think you need to do a if else here. Just start as the validation to true and then if you have additional panels do the the validation to change that boolean 

```javascript
var isValidAdditionalSign = true;
...
if (!_.isEmpty(additionalPanelsProps.values)) {
...
}
```

UI/src/assetTypeConfiguration.js
marquesfm: Done

UI/src/assetTypeConfiguration.js
marquesfm: Done

PR #1579, opened 2020-09-02
DROTH-2529 (U.S:2454) Extraction board the button to show log info doesnt work in case of error

ricardofmsimoes:
- Fix the button to show the log info doesn't work in case of error

### DIFF
--- a/UI/src/view/csvReportsWorkList.js
+++ b/UI/src/view/csvReportsWorkList.js
@@ -207,7 +207,7 @@
             '</div>'
         );
 
-        $('#work-list .work-list').html(me.workListItemTable(result));
+        $('#work-list .work-list').html(me.workListItemTable());
 
         populatedMultiSelectBoxes();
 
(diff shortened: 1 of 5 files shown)


marquesfm (approved):

UI/src/view/csvReportsWorkList.js
```diff
@@ -388,19 +378,10 @@
...
-          '</header>' +
-          '<div class="error-list">' +
-          '</div>'
-      );
-      jobView.find('.header-link').on('click', function(){
-        showImporter();
-      });
-      $('.error-list').html(job.content);
+      return  '<header>' +
+                '<h2> CSV-eräajon virhetilanteet: ' + job.fileName + '</h2>' +
```
sasuolanderSito, line 382: In requirement is Finnish marked as an only required language? There is no need for multilanguage support?

ricardofmsimoes: @sasuolanderSito 
No, there is no multiple languages support.
It was never a requirement. And it is only used by finnish administrative authority (or something similar).

PR #1587, opened 2020-09-23
Droth 2541 road marking service

sasuolanderSito:
okey this is just draft request, I have been rather slow to write code because thinking and analysing has taken a lot of time,

One thing I was thinking that default linearasset GET method can be used to retrieve these to map,

to add multiple link I used list of newlink, same for update method, might be better to update only one at time


digiroad2-api-oth/src/main/scala/fi/liikennevirasto/digiroad2/Digiroad2Api.scala
```diff
@@ -2354,4 +2357,116 @@ class Digiroad2Api(val roadLinkService: RoadLinkService,
...
+    var test ="test"
+    test
+  }
+  //add
+/*{
+    "typeId":460,
+    "lengthOfRoadAxis":{
+        "lengthOfRoadAxi":{"typeId":460 "asset":["newLinearAsset"]}
+    }
+}*/
```
marquesfm, line 2373: This is needed? If not, please deleted.

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/linearasset/LengthOfRoadAxisService.scala
```diff
@@ -0,0 +1,86 @@
...
+   username: String
+  ) = {
+   val modifiedElement=
+     withDynTransaction{
+       for (item <- listOfAsset) yield {super.updateWithoutTransaction(item.ids, item.value, username)}
+     }
+    eventBusImpl.publish("LengthOfRoadAxisService:update",modifiedElement.toSet)
+    modifiedElement
+  }
+}
```
marquesfm: The idea was to use the current methods already existent on DynamicLinearAssetService, maybe we need to override the create, update and delete methods since on this user story it's needed to work with more than one asset at the same type.

But nothing too much, like receive multiple and then call the current methods multiples times

digiroad2-oracle/src/test/scala/fi/liikennevirasto/digiroad2/service/linearasset/LengthOfRoadAxisServiceSpec.scala
```diff
@@ -0,0 +1,152 @@
+package fi.liikennevirasto.digiroad2.service.linearasset
```
marquesfm, line 1: Adjust unit tests according to the suggestions above

marquesfm (changes requested):

ricardofmsimoes (commented):
Added some comments

digiroad2-oracle/src/main/scala/fi/liikennevirasto/digiroad2/service/linearasset/LengthOfRoadAxisService.scala
sasuolanderSito: In what way old one allows only one? If in frontend we create arrays of multiple assets of one type, we can run multiple post request to POST:/linearasset/ endpoint.  

Or was the idea that one roadlink can have multiple values. In one roadlink there might be K1 and K2 mark the same time for example.

PR #1600, opened 2020-11-18
Revert "DROTH-2563 Green color option for additional plates"

sasuolanderSito:
Reverts finnishtransportagency/digiroad2#1585
if we need this change in future, we need to run revert of revert pull request.


SampsaHyvamaki (commented):
kaikki ok

PR #1604, opened 2020-11-23
DROTH-2614 Edited picture with missing lower part by adding lower par…

rene-sitowise:
…t to it.

Edited old picture to 
image


sasuolanderSito (commented):
ok

sasuolanderSito (approved):

PR #1613, opened 2020-12-14
Deleted visualization of info value of traffic sign

rene-sitowise:
Deleted visualization of info value of traffic signs.


sasuolanderSito (approved):
ok

annarik (approved):

PR #1615, opened 2020-12-17
Revert "DROTH-2598 Changed value on mainSignText field to same as additionalI…"

rene-sitowise:
Reverts finnishtransportagency/digiroad2#1611


sasuolanderSito (approved):
vaikuttaa ok

PR #1624, opened 2021-02-09
Fix Akka Logging

AnttiJokela:
Exceptions in async processing code were lost. Added configuration to redirect the error messages to application log file for further analysis.
Also Improved the logging of missing roadlink while hanlding side code change.


sasuolanderSito (approved):
ok

PR #1655, opened 2021-03-26
 DROTH-2725 fix invalid insert

sasuolanderSito:
when deleting masstransitstop it ID cannot be saved to vallulog table


sasuolanderSito:
testit ok

rene-sitowise (approved):

PR #2231, opened 2022-09-20
DROTH-3418 bigger compute type for dev and test codeBuild

anttiahopeltositowise:
Tehoja nostettu dev ja test codebuild


sasuolanderSito (approved):
ok

PR #2328, opened 2022-11-23
DROTH-2640 Allow five additional panels for traffic sign

reetta-kivisto:
Dev ok. Testattu UI, csv import, Kalpa API ja Change API.


anttiahopeltositowise (approved):
ok

PR #2519, opened 2023-07-12
Droth 3769 point assets end measure to null

MJKaakinen:
-Removed instances where a point asset's endMeasure is set to value
 -Modified LRMPosition case class into PointAssetLRMPosition


anttiahopeltositowise (approved):
OK

sasuolanderSito (approved):
ok

PR #2599, opened 2023-08-29
Revert "Revert "Droth 3744 enable tram stops on pedestrian links""

sasuolanderSito:
revert Reverts finnishtransportagency/digiroad2#2596


SWJPTulijoki (approved):
ok

PR #2911, opened 2024-02-13
DROTH-3885 Stop using recalculate bearing, use roadlink bearing if wi…

anttiahopeltositowise:
QA:lle testattavaksi


SWJPTulijoki (approved):
ok

PR #2919, opened 2024-02-14
 DROTH-3954 filter changes where municipality code is null

sasuolanderSito:
jos kuntakoodi on null, ohitetaan se


SWJPTulijoki (approved):
ok

PR #3235, opened 2024-07-19
Droth 4036 fix remaining aws inspector warnings after removing dev dependencies from build

mvrkus-sitowise:
Toimii devillä


SWJPTulijoki (approved):
ok

PR #3259, opened 2024-08-19
DROTH-4107 Added new obstacle types to be accepted in csv imports.

mvrkus-sitowise:
Testaukseen qa:lle


anttiahopeltositowise (approved):
ok

PR #3325, opened 2024-10-22
DROTH-4123 samuutus asset expire bug

mvrkus-sitowise:
Fix change set size in RoadLinkChangeClientSpec


anttiahopeltositowise (approved):
ok

PR #3393, opened 2025-02-19
 DROTH-4255 Display historylink municipalityCode as 0 when null to retain uniform practice across process

MJKaakinen:
Testattu devillä, toimii


anttiahopeltositowise (approved):
OK

PR #3402, opened 2025-02-26
DROTH-3172 change roadlink fetching function to one that includes ove…

mvrkus-sitowise:
…rride adminclasses.

Toimii devillä


anttiahopeltositowise (approved):
ok

PR #3485, opened 2025-07-28
DROTH-4191 Fix bug on speedLimitLayer. Wrong parameter was given to f…

anttiahopeltositowise:
Dev testaus OK


sasuolanderSito (approved):
ok

PR #3582, opened 2025-11-14
DROTH-4468 Avoid MassQuery when fetching point assets because it's re…

LauriLitmanen:
…ally slow at the moment.


sasuolanderSito (approved):
ok

